PR #1, opened 2024-03-28
initial commit with probe matcher script

kaitlyn-sharo:
See https://github.com/NextCenturyCorporation/itm-scenario-validator/pull/22 - moved to ingest repo


brianpippin (approved):

joshuale-caci (changes requested):
I noticed that there are some files already in the metrics data folder with this commit. Should we add them to the .gitignore or are these files necessary.

kaitlyn-sharo:
> I noticed that there are some files already in the metrics data folder with this commit. Should we add them to the .gitignore or are these files necessary.

From my understanding, @brianpippin wanted them here, as a place to keep the clean dataset. The data doesn't download in the right format from confluence and ends up with many extra files (files without actions, tutorial files, etc.), so this is one place where we can find only the valid, clean data from the sim

joshuale-caci (approved):

brianpippin:
> > I noticed that there are some files already in the metrics data folder with this commit. Should we add them to the .gitignore or are these files necessary.
> 
> From my understanding, @brianpippin wanted them here, as a place to keep the clean dataset. The data doesn't download in the right format from confluence and ends up with many extra files (files without actions, tutorial files, etc.), so this is one place where we can find only the valid, clean data from the sim

FYSA - Ideally we are getting these files from S3 or some other resource in the future but for now, this is the best option we have I believe.

PR #3, opened 2024-03-28
Add all ADM history data and yaml scene data to database

brianpippin:
This gets all the metrics evaluation data into the database, however the dashboard DOES NOT look right.  There is a corresponding Dashboard fix to update the labels and to group the datasets by evaluation that will be coming and I am working on.


kaitlyn-sharo (commented):
I can't run this. It keeps saying something about authentication. What are the instructions for running?

kaitlyn-sharo:
> I can't run this. It keeps saying something about authentication. What are the instructions for running?

Actually, I think it was that mongo was pointed to the wrong port now that we've had to change it

brianpippin:
> > I can't run this. It keeps saying something about authentication. What are the instructions for running?
> 
> Actually, I think it was that mongo was pointed to the wrong port now that we've had to change it

Yes we might need to update the port here too if we are changing this permanently.

kaitlyn-sharo (approved):
Seems to work with the new dashboard page!

PR #5, opened 2024-04-05
updated data format; added 4/5 data collection

kaitlyn-sharo:
Organized data like the new sim format, where there's only one parent folder that contains the csv and json. CSVs are now named correctly (with the participant id). Made a minor code change in the main function to handle this new organization.

Added OSU data from 4/5 collection


dereknop (approved):
Probe matcher still worked for me. The data format looks good to me. Looking over the data, I did find it weird that participant 2024209's adept urban run only has one probe in its alignment source. Is this just because that participant only partially completed that scenario?

kaitlyn-sharo:
> Probe matcher still worked for me. The data format looks good to me. Looking over the data, I did find it weird that participant 2024209's adept urban run only has one probe in its alignment source. Is this just because that participant only partially completed that scenario?

Yes, some of the data has been lost. I believe 209 is one where Jennifer went through the csvs to find as much data as she could for the probes and I recreated the json based on that. Since adept probes are voice-based and not recorded in the csv, those are a lot harder to find than the soartech probes

brianpippin (approved):

PR #6, opened 2024-04-09
ITM-334: Update human sim records with sim order and average

brianpippin:
Must run this to get all columns in Kaitlyn's ITM-324 PR that I also worked on.


dereknop (approved):
I ran into no issues running the updated script and the data is reflected in the dashboard

kaitlyn-sharo (approved):

PR #7, opened 2024-04-10
ITM-338:  Script to remove old ADM runs and add new ADM runs

brianpippin:
Notes:  You should be able to ignore the files in metrics-adm-data for review.  I only copied the new files in and deleted the old files.

Also deployment_script.py has updates to run multiple scripts currently, that is because the deployment script hasn't run on production yet.  Ideally in the future, once we deploy to 'itm-ingest' main, Jenkins will have a job that will pick that up, and automatically run the deployment_script for us.


kaitlyn-sharo (approved):
Worked for me!

dereknop (approved):
No issues running deployment_script.py. Verified mongo was updated with new data

PR #8, opened 2024-04-11
ITM-178: Add script to add admin users to production

brianpippin:
Two Notes here:

1.  If you want to test the corresponding UI ticket that goes with this, you might want to add your local username to this script before you run it.
2. We are still running all the old scripts because we haven't pushed ingest to main.  We want one script call that will run all the scripts on production, ideally in the future, we merge one script to main at a time.


dereknop (approved):
Worked after adding my username 'dereknop' to the array

kaitlyn-sharo (approved):
I saw the non-admin dashboard after running the first time, then added my name to the array and saw the admin side. Looks good to me

PR #10, opened 2024-04-15
Text based alignment scores script refactored to run with the rest of…

dereknop:
… deployment scripts

To test run through the text-based scenarios on the dashboard and then run deployment_script.py here. Check new Mongo documents from completing text-based scenarios to make sure alignment scores are attached.

After battling with CORS errors for way too long, I decided to refactor this script to go with the rest of the deployment scripts.


deployment_script.py
```diff
@@ -5,11 +5,11 @@
 from scripts._0_0_5_remove_old_adm_data import remove_old_adm_records
 from scripts._0_0_6_set_admin_users import add_admin_user_role
 from scripts._0_0_7_update_scenario_index import update_scenario_index
-
```
brianpippin, line 8: The other scripts have already run on production so we should remove anything 0.0.7 or earlier.

deployment_script.py
```diff
@@ -44,6 +44,7 @@ def main():
         add_eval_3_adm_data(mongoDB)
         add_admin_user_role(mongoDB)
         update_scenario_index(mongoDB)
```
brianpippin, line 46: We should remove this other calls because those scripts have run, we should also probably remove the probe matcher script from running now that it has run successfully.

textbased-configs/adeptDesertConfig.json
brianpippin, line 1: I think we should rename this folder so we know it is the metrics text based configs

kaitlyn-sharo (approved):
Awesome, this'll make everything so much cleaner. Probably needs a corresponding cleanup branch on the dashboard too, just so we don't have this code in two places

brianpippin (approved):
Ran with no issues.  Worked great.

PR #12, opened 2024-04-16
Itm 346

dereknop:
This attaches highAlignmentData and lowAlignmentData for all of the text-based scenario results. Also clears out the pre-existing 'alignmentData' field from the results. 

Note: This change will break any graphql query pulling alignmentData from mongo documents. That will need to be substituted to pull from these two new fields.


brianpippin:
We are definitely going to need a corresponding UI fix for this before we can run this on production.

kaitlyn-sharo (approved):

brianpippin (approved):
Code ran and worked well, I am not sure we need the print statements though, I think they are good for debugging but not sure we need them on the production run.

PR #13, opened 2024-04-24
Itm 349

dereknop:
`python deployment_script.py`

Run this script and then test the survey results page. The page should now have scenario names instead of scenario indexes. I.e Soartech Submarine instead of "scenario 1". Also, omnibus results were only showing the comparison page between two omnibus and not questions for each omnibus separately. There should now be three sections under omnibus survey results, each omnibus individually and then the comparison.

### DIFF
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -1,6 +1,5 @@
 from pymongo import MongoClient
 import os
-from scripts._0_0_8_add_textbased_alignment import add_textbased_alignments
 from scripts._0_0_9_add_scenario_names import add_scenario_names
 VERSION_COLLECTION = "itm_version"
 
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -1,10 +1,11 @@
 from pymongo import MongoClient
 import os
 from scripts._0_0_8_add_textbased_alignment import add_textbased_alignments
+from scripts._0_0_9_add_scenario_names import add_scenario_names
 VERSION_COLLECTION = "itm_version"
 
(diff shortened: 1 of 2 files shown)


deployment_script.py
```diff
@@ -1,10 +1,11 @@
 from pymongo import MongoClient
 import os
 from scripts._0_0_8_add_textbased_alignment import add_textbased_alignments
+from scripts._0_0_9_add_scenario_names import add_scenario_names
```
nextcen-dgemoets, line 4: I had to remove the import of `add_textbased_alignments` due to the character encoding issue we encountered when trying to get my dashboard running.  The import is no longer needed here anyway.   I assume it's a separate ticket to take the actions we discovered when trying to get my dashboard working.

scripts/_0_0_9_add_scenario_names.py
```diff
@@ -0,0 +1,41 @@
...
+                # add field to existing omnibus results for surveyResults page
+                if scenario_index in (9, 10):
+                    page['pageType'] = 'singleMedic'
+                
+
+        survey_results_collection.update_one(
+                {'_id': result['_id']},
+                {'$set': result}
+            )
+
```
nextcen-dgemoets, line 39: I'd suggest some nice little "Update complete" output at the end of the script.

nextcen-dgemoets (approved):
All of the new stuff was there and everything worked and looked good.

I had a couple of minor suggestions you can take or leave.

brianpippin (approved):

PR #14, opened 2024-04-29
updates to readme based on issues Darren ran into

dereknop:
Updated documentation and issue darren ran into with text-based alignments script

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ To run the probe matcher, execute the following command:
 ```
 python3 probe_matcher.py -i [path_to_directory]
 ```
-Ensure that the path leads to the Unity Simulator data. The format should be: top-level -> id -> id -> json, with csvs directly under top-level. As the Unity file output organization format changes, the way of accessing this data may also change.
+Ensure that the path leads to the Unity Simulator data. The format should be: top-level -> id -> id -> json, with csvs directly under top-level. As the Unity file output organization format changes, the way of accessing this data may also change. For the metrics evaluation data, use the folder in this repo 'metrics-data'.
 
 ### Output
 The probe matcher will output two json files for each input json found: one for soartech and one for adept. If SEND_TO_MONGO is true, these json files (and the raw jsons found) will be sent to the mongo database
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,4 +61,4 @@ To utilize this function, you will need to manually run the deployment script.
 We will eventually have a Jenkins job that will run the script whenever ingest is updated.
 Also if you would like to run a new script you will need to modify the script that is being called in deployment script and bump the version.
 
-The script will only run if the version is a newer version of the db, the script also create the collection for the version if it doesn't exist yet.
+The script will only run if the version is a newer version of the db, the script also create the collection for the version if it doesn't exist yet. Set your .env file to point to the instances of the TA1 servers you have running locally.
```
nextcen-dgemoets, line 64: I was able to use `https://darpaitm.caci.com/soartech/` and `https://darpaitm.caci.com/adept/`, which is handy if you haven't or don't want to run local TA1 servers.  Of course it requires that those servers are running...

nextcen-dgemoets (commented):
I also had to update `deployment_script.py` to add invocation of six scripts (0_0_2 to 0_0_7), including all of the imports at the top of the file.  Maybe we add the includes in the file and add the invocations commented out to facilitate a new installation without breaking current executions of the deployment script?

README.md
dereknop: I think we orignially had those in the .env but they didn't work on production. I can add them back in just commented out for use locally.

nextcen-dgemoets: Oh right-- yes it should definitely be commented out.  It should work on production by default.

dereknop:
> I also had to update `deployment_script.py` to add invocation of six scripts (0_0_2 to 0_0_7), including all of the imports at the top of the file. Maybe we add the imports in the file and add the invocations commented out to facilitate a new installation without breaking current executions of the deployment script?

If someone grabs a copy of the production database from AWS, they wouldn't need to run those scripts because they will have already been run on production, with the exception of script 0_0_6 which they will need to run to set themselves as an admin locally. In the updated documentation for the dashboard, I made note that they will have to run that script under the Admin Account section.

nextcen-dgemoets:
> If someone grabs a copy of the production database from AWS, they wouldn't need to run those scripts because they will have already been run on production, with the exception of script 0_0_6 which they will need to run to set themselves as an admin locally. In the updated documentation for the dashboard, I made note that they will have to run that script under the Admin Account section.

Okay-- so I had to run them because I got my DB dump from you and not from production / AWS?  In other words, if my case was a case that won't need to be repeated, then it makes sense not to include them in the script.  If somebody might need them, then they could be left in uncommented.

nextcen-dgemoets (approved):

brianpippin (approved):

PR #15, opened 2024-04-29
Fix participant IDs that are incorrect and not matching

brianpippin:
So I considered making the version 0.0.10 but because of string comparison that is less that 0.0.9 so I went with 0.1.0.  I still think this allows us up to so many scripts for the life of the program we should be good.


scripts/_0_0_10_fix_participant_id.py
```diff
@@ -0,0 +1,24 @@
+def fix_participant_id(mongoDB):
```
nextcen-dgemoets, line 1: If the DB version is 0.1.0, presumably this filename should be `_0_1_0_fix_participant_id.py`?

deployment_script.py
```diff
@@ -35,6 +36,7 @@ def main():
         print("New db version, execute scripts")
 
         add_scenario_names(mongoDB)
```
nextcen-dgemoets, line 38: I'm unclear on when these lines are removed (or commented out) from the deployment script.  Is there a rule of thumb?  And does it matter-- that is, are the scripts assumed to be idempotent (i.e., that re-running them has no effect)?

Part of the reason I ask is because I've already run `add_scenario_names`.

dereknop (approved):
No issues with the script. Was able to confirm there are four files for each of those updated participantID's in the userScenarioResults collection after running

nextcen-dgemoets (approved):
Everything looked good.  I verified (eventually) that those records with spurious IDs were in the database prior to running the script, and after running the script, they were not there, and did not appear in the dashboard either.

I did see a spurious Participant ID ("der123"), but it turned out to be a local artifact.  It was NOT on production.

PR #16, opened 2024-04-30
itm-370

dereknop:
https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-370

A short script that cleans out extraneous data in the text-based results collection. This is all data that didn't have participantID's attached to it and was from testing before the text-based scenarios were complete (all dummy data). None of this data was being pulled to the dashboard so there will be no visual difference there.


nextcen-dgemoets (approved):
Looks good.  After running the script, the number of records in my `userScenarioResults` table decreased from ~128 to 96.  And everything looked good in the dashboard.

PR #18, opened 2024-06-14
Start adding survey config to mongo

kaitlyn-sharo:
Run `python3 delegation_tool.py` 

This is a command-line tool designed to help add survey configs to mongo as well as update them. Play around with some of the different options. Then run 3 times.

Time 1:
Version: `1.0`
Choice: `is`
`surveyConfig1x.json`
`y`
Choice: `s`
`y`
`q`

Repeat for version 2.0 and 0.0, with surveyConfig2x.json and test_config.json, respectively

This will go through future iterations as needed for additional editing of configs. This is just the bare bones to get us started

### DIFF
--- a/delegation_tool.py
+++ b/delegation_tool.py
@@ -179,7 +179,8 @@ class DelegationTool:
         Adds the initial survey base to the database
         '''
         if not is_new:
-            overwrite = input("This will overwrite all previous delegation survey data in the database. Are you sure you want to do this? (y/n) ")
+            LOGGER.log(LogLevel.WARN, "This will overwrite all previous delegation survey data in the database. Are you sure you want to do this? (y/n) ")
+            overwrite = input("")
         if is_new or overwrite.strip() in ['y', 'Y']:
             self.survey = {
                 "title": "ITM Delegation Survey",
@@ -280,7 +281,7 @@ class DelegationTool:
         Move a page with page_name from a json file to the survey 
         '''
         self.cur_import_file = json_dest
-        f = open(json_dest, 'r')
+        f = open(json_dest, 'r', encoding='utf-8')
         data = json.load(f)
         if 'pages' not in data:
             LOGGER.log(LogLevel.WARN, "Error: json needs a 'pages' key at the top level to run 'import_page_from_json")
@@ -328,10 +329,10 @@ class DelegationTool:
         '''
         Places an entire json full of survey data (found at path) in the survey
         '''
-        LOGGER.log(LogLevel.WARN, "This overwrites all current survey data. Are you sure you want to continue? (y/n)")
+        LOGGER.log(LogLevel.WARN, "This will overwrite all current survey data. Are you sure you want to continue? (y/n)")
         resp = input("")
         if resp.strip() in ['y', 'Y']:
-            f = open(path, 'r')
+            f = open(path, 'r', encoding='utf-8')
             self.cur_import_file = path
             data = json.load(f)
             self.survey = data
@@ -373,7 +374,13 @@ if __name__ == '__main__':
             if resp == 's':
                 tool.push_changes()
             if resp == 'v':
-                print(json.dumps(tool.survey, indent=4))
+                survey_copy = copy.deepcopy(tool.survey)
+                for p in survey_copy['pages']:
+                    for e in p['elements']:
+                        if 'patients' in e:
+                            for pat in e['patients']:
+                                pat['imgUrl'] = 'Image data omitted from printout.'
+                print(json.dumps(survey_copy, indent=4))
             if resp == 'q':
                 if len(tool.changes_summary) > 0:
                     LOGGER.log(LogLevel.WARN, "Changes are not saved. Are you sure you want to quit? (y/n)")
--- a/delegation_tool.py
+++ b/delegation_tool.py
@@ -1,0 +1,403 @@
+
+from pymongo import MongoClient
+from decouple import config 
+import json, time, copy
+from logger import LogLevel, Logger
+
+'''
+The delegation survey python tool adds the delegation survey questions to the database 
+and can modify, add, order, and delete questions from the survey config.
+'''
+LOGGER = Logger('Delegation Tool')
+
+
+class SurveyElement:
+    def __init__(self, minimum=None, visible_if=None, input_type=None, question_choices=None,
+                 decision_makers=None, actions=None, is_required=None, show_other=None,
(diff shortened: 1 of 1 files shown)


dereknop (commented):
The three different configs were added properly to a new `delegationConfig` collection. One thing I noticed is that the `version` field of the test file is still `2`. I don't know if this has any impact on changes to the dashboard so figured I'd mention it.

dereknop (approved):

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+                page_name = input("Enter the name of the page to import: ")
+                ind = input("At what index should this page be inserted? (enter to append) ")
+                tool.import_page_from_json(path if path != '' else tool.cur_import_file, page_name, int(ind) if ind != '' else None) 
+            if resp == 'dp':
+                page_name = input("Enter the name of the page to delete (only the first page found with this name will be deleted): ")
+                tool.delete_page(page_name)
+            if resp == 's':
+                tool.push_changes()
+            if resp == 'v':
+                print(json.dumps(tool.survey, indent=4))
```
nextcen-dgemoets, line 376: This isn't very useful as-is, because the images just spew text on the screen.  As an enhancement request (either for now or down the line), before printing, it would be nice to remove `imgUrl` or similar keys from the result of `json.dumps` or replace their values with a placeholder, like `Image data omitted.`

nextcen-dgemoets (commented):
Things seem to work well, and I love the user interface.
After I made selections, I could see changes reflected in the DB.

I just had a few minor questions/suggestions.

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            LOGGER.log(LogLevel.CRITICAL_INFO, "Okay, not updating survey.")
+
+
+    def import_page_from_json(self, json_dest, page_name, index=None):
+        '''
+        Move a page with page_name from a json file to the survey 
+        '''
+        self.cur_import_file = json_dest
+        f = open(json_dest, 'r')
```
nextcen-dgemoets, line 283: I forgot to mention-- both here and the other `open` call, for this to work on Windows, you need to add `encoding="utf-8"` to the call.  Derek and I ran into this in `_0_0_8_add_textbased_alignment.py`.

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+                if not update:
+                    self.changes_summary.append(f"Deleted page '{page_name}'.")
+                return
+            except:
+                LOGGER.log(LogLevel.WARN, f"Page '{page_name}' was not found. Could not {'delete' if not update else 'update'}.")
+        else:
+            LOGGER.log(LogLevel.INFO, f"Not deleting page '{page_name}'")
+
+
+    def add_element(self, page_name, survey_el, index=None):
```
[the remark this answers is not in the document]
kaitlyn-sharo: Yes, I was originally testing and didn't have the command line tool, so these did work (they were actually the first functions I made), but it was longer than just importing. I imagine they will still be used later on though.

delegation_tool.py
kaitlyn-sharo: Oh that's interesting. I don't actually store the image data in mongo like that (it is pulled out and there's just a placeholder there) so yes, I should not be printing it. Thanks for catching that!

delegation_tool.py
kaitlyn-sharo: That is good to know, thanks! I'll fix that too

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+        if not found_version:
+            self.changes_summary.append(f"Added new survey version '{version}' to database")
+            self.initialize_survey(version, True)
+            self.survey_mongo_collection.insert_one({'_id': obj_id, 'survey': self.survey})
+        
+        if 'survey' in self.survey:
+            self.survey = self.survey['survey']
+
+
+    def add_page(self, page_name, scenario_name='', page_type='', scenario_index=None, index=None):
```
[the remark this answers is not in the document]
kaitlyn-sharo: Yes

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+                return page
+        return False
+
+
+    def initialize_survey(self, version, is_new):
+        '''
+        Adds the initial survey base to the database
+        '''
+        if not is_new:
+            overwrite = input("This will overwrite all previous delegation survey data in the database. Are you sure you want to do this? (y/n) ")
```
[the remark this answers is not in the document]
kaitlyn-sharo: Good catch, again. Thanks!

delegation_tool.py
```diff
@@ -0,0 +1,387 @@
...
+        if 'survey' in self.survey:
+            self.survey = self.survey['survey']
+
+
+    def add_page(self, page_name, scenario_name='', page_type='', scenario_index=None, index=None):
+        '''
+        Adds a new page to the survey
+        '''
+        # check for existing page with name before adding
+        if not self.handle_existing_page(page_name):
```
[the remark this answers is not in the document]
kaitlyn-sharo: No, because handle_existing_page does that already :)

delegation_tool.py
kaitlyn-sharo: I had to pipe this out to a file in order to see it all. In future iterations, I may look into automatically printing with some sort of more/less command, and I plan to add a print_to_json feature as well, that way we can see what's in the database easily.

kaitlyn-sharo:
@brianpippin How do you want me to organize this script with all the other scripts in here? Should I keep it where it is or move it somewhere else? There is going to be one other script for grabbing the adm data and formatting it for the survey, so I'm thinking both scripts and all the config files should at least be together wherever we want them

brianpippin:
> @brianpippin How do you want me to organize this script with all the other scripts in here? Should I keep it where it is or move it somewhere else? There is going to be one other script for grabbing the adm data and formatting it for the survey, so I'm thinking both scripts and all the config files should at least be together wherever we want them

So I think it depends on how often this script will need to be run.  Ideally for the Dry Run Evaluation, I would like some thing to run configured by AWS that when an ADM output file gets uploaded to S3 (which already happens), it triggers an automated ingest to mongo.  If this script would need to be run every time we did that, then it should be somewhere where we can use it for that process.  Ideally it would not be run post data collect anymore but during data collect.

nextcen-dgemoets (approved):
All of the changes look good to me!

PR #22, opened 2024-08-13
ITM-552: Reduced text based images sizes to 25percent of original

phile-caci:
The large images were causing an impact on loading time.  This ticket just reduces the image sizes to speed up the loading time as a quick fix.  Images were reduced to 25% of original. The image size/depth are probably much larger than what we need them for.  

(new image are already ingested in prod)

All new images are about  6mb from 230mb
reviewtextscenarios load time was 50 sec, 350mb, now it's 8 secs 53mb, still kinda large/long...

Further optimization is still needed.

Just some quick profiling, the top 4 slow areas we should look at:
1. programQuestions is large also, 153mb, 23secs
2. Reviewtextscnarios still large 53mb, 8secs
3. humanResults 3.2mb, 478ms
4. aggregateResults, 2.7mb, 400ms
image


dereknop (approved):

nextcen-dgemoets:
What do I need to do to test this?  Do I need to delete any existing tables?  Do I need to re-run `deployment_script.py`?

brianpippin (approved):
Good idea on the file sizes, if a ticket doesn't exist to look at those other calls we should create a ticket for that, so we can try to optimize load times after all the other items are done.

phile-caci:
> Good idea on the file sizes, if a ticket doesn't exist to look at those other calls we should create a ticket for that, so we can try to optimize load times after all the other items are done.

Thanks Brian, I'll create a ticket now...

PR #24, opened 2024-08-19
Itm 551

dereknop:
[ITM-551](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-551)

Depends on corresponding [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/compare/development...ITM-551)

Run `python3 deployment_script.py` from inside the venv. This will replace the existing relevant collections so nothing will need to be manually deleted in MongoDB.

Fixes:

- Updated ST files with correct unstructured text for actions as well updates to character vitals
- Descriptions for probes being pulled from probe config for ST. Manually inserted a special character to split unstructured value from Adept scenes to display only new relevant info.
- Keep track of more mission details for ST (see more info modal in UI)
- Keep track of when vitals shouldn't be visible (if assessing the patient is an action in the action mapping, vitals should show as `unknown`)
- Logic to account for transition scenes (only applies to Adept, see modal in UI)
- Refactored logic for conditional rendering of pages in Adept scenarios 
- Support for 'why probes' in Adept scenarios. Certain action choices will prompt for an explanation in a separate question, while other answers may not. See the first probe of MJ2.

[ITM-551]: https://nextcentury.atlassian.net/browse/ITM-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dereknop:
I should be more explicit that the special character inserted was a $ for adept

kaitlyn-sharo (approved):

brianpippin:
> I should be more explicit that the special character inserted was a $ for adept

Would it make more sense that instead of adding one character you added something extremely specific like <text_scenario_line_break> or something just in case a scenario in the future actually has a dollar sign?

brianpippin:
Everything looks good on my end however it looks like there are merge issues, I'd like to test it again once those have been resolved for a sanity check.

dereknop:
> > I should be more explicit that the special character inserted was a $ for adept
> 
> Would it make more sense that instead of adding one character you added something extremely specific like <text_scenario_line_break> or something just in case a scenario in the future actually has a dollar sign?

Fair point, that would be better. I will change that.

nextcen-dgemoets:
Would love to review this, but can't until IT lets me upgrade my WSL...

phile-caci (approved):
LGTM, ran the deplotment_script, no issues.

PR #25, opened 2024-08-21
created updatedOrderLog field

dereknop:
This matches the medic names to what was recorded in the order log in the surveyResults collection. We record this new array in the updatedOrderLog field just in case because I did not want to clobber the existing data.

To test `python3 deployment_script.py`

Then look in MongoDB to see that the survey results for survey version 3 have the new field

Look in the [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/compare/updatedOrderLogs?expand=1) to make sure you can also see this field when downloading the excel file for survey version 3.


phile-caci (approved):
LGTM:

image

kaitlyn-sharo (approved):

PR #27, opened 2024-08-23
Itm 566 ingest eval4 scenarios

phile-caci:
Eval 4 Scenario ingest for use in UI ADM Data.

You'll need to: 

Once data gets populated you should be able to click through all the selections on http://localhost:3000/results:
image

### DIFF
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -3,11 +3,11 @@ from decouple import config
 import os
 from text_based_scenarios.convert_yaml_to_json_config import main as generate_textbased_configs
 from text_based_scenarios.add_patient_pictures import main as add_pictures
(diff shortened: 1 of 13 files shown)


deployment_script.py
```diff
@@ -3,11 +3,11 @@
 import os
 from text_based_scenarios.convert_yaml_to_json_config import main as generate_textbased_configs
 from text_based_scenarios.add_patient_pictures import main as add_pictures
```
brianpippin, line 5: Don't think we need lines 4 and 5 here any more.

dereknop (approved):

deployment_script.py
phile-caci: removed, thanks!

PR #33, opened 2024-09-06
supplies fix

dereknop:
To test,

http://localhost:3000/review-text-based

I found a bug earlier today where, after the supplies were updated, that update didn't persist in the Adept scenarios. After the scene where new supplies were introduced, it would revert to the scenario's initial supplies. That should be fixed now. MJ4 and MJ2 have supply updates that should now be properly persisted.

### DIFF
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -5,7 +5,7 @@ VERSION_COLLECTION = "itm_version"
 MONGO_URL = config('MONGO_URL')
 
 # Change this version if running a new deploy script
-db_version = "0.1.11"
+db_version = "0.2.0"
 
 
 def check_version(mongoDB):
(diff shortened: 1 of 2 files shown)


deployment_script.py
```diff
@@ -1,11 +1,11 @@
 from pymongo import MongoClient
 from decouple import config
-from text_based_scenarios._0_1_10_participant_log import main as ingest_participant_log
+from text_based_scenarios.convert_yaml_to_json_config import main as generate_text_based_configs
 VERSION_COLLECTION = "itm_version"
 MONGO_URL = config('MONGO_URL')
 
 # Change this version if running a new deploy script
-db_version = "0.1.10"
+db_version = "0.1.11"
```
brianpippin, line 8: This won't work on the version check, the strings won't think it is larger, we had to do a hack for the 0.1.10, make this one 0.2.0 (my script is going to be 0.2.1)

kaitlyn-sharo (approved):

brianpippin (approved):

PR #36, opened 2024-09-10
Adept location supply change

dereknop:
Adept treatments should no longer say 'Treated passerby with a tourniquet to leg'. Instead will just say 'Treated passerby'. There is an edge case in MJ2 that will still say the quantity of gauze used to treat the victim or shooter. 

`cd delegation_survey`
`python3 dre_adms_for_delegation.py`
`python3 update_survey_config.py`
`enter`
`3`
`y`
`y`

http://localhost:3000/review-delegation

### DIFF
--- a/delegation_survey/dre_adms_for_delegation.py
+++ b/delegation_survey/dre_adms_for_delegation.py
@@ -812,8 +812,10 @@ def remove_location_and_supply(scene):
     for i in range(len(scene['actions']) - 1):
         if 'Question: How much gauze do you plan to use on the shooter?' in scene['actions'][i]:
             scene['actions'][i+1] = scene['actions'][i+1].split('on')[0].strip()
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
Looks good.

In the edge case, I wonder if "Treat Shooter with hemostatic gauze" should say, "Treat Shooter with 1 hemostatic gauze" to go along with "Treat Victim with 4 hemostatic gauze."

brianpippin:
> Looks good.
> 
> In the edge case, I wonder if "Treat Shooter with hemostatic gauze" should say, "Treat Shooter with 1 hemostatic gauze" to go along with "Treat Victim with 4 hemostatic gauze."

If this an easy change we should implement it, if it's not a 5-10 minute change we should not.

kaitlyn-sharo (approved):

PR #37, opened 2024-09-12
Update Post Questions per Jennifer's instructions

brianpippin:


nextcen-dgemoets:
I re-ran `update_survey_config.py` but still got the old questions.  I rebuilt my container and still got the old questions.
LMK what other step I was supposed to take to see those questions gone and I'll try a third time.

dereknop:
re run `python3 update_survey_config.py`
`enter`
`3`
`y`
`y`

If you want to see that the page has changed without clicking all the way back to it add this to line 859 of survey.jsx
`console.log(pages[pages.length - 1])`

phile-caci (approved):
Thanks Derek/Brian, checked against Jennifer's doc, looks good

PR #38, opened 2024-09-13
Final dre changes

dereknop:
Run `python3 deployment_script.py` from root of directory.

Then 
`cd delegation_survey`
`python3 dre_adms_for_delegation.py`
`python3 update_survey.py`
`enter`
`3`
`y`
`y`

Text based scenarios should no longer have titles on the page like "probe 2.1"
Same goes for delegation survey no longer having scenario names displayed to participants
Question in post scenario measures page had answer key change


phile-caci (approved):
LGTM...

Survey - Title (scenario id is gone)
Text - probe titles are gone
Survey post question has changed

brianpippin (approved):

PR #39, opened 2024-09-17
Optimize Initial Startup

phile-caci:
Reducing image size... This should drop the initial size  (65mb) of the GetConfigs call by 23mb.

Reingest with:
python3 update_survey_config.py
enter
1
n
n
n
n
y
y (edit: missing last y)

LMK if I missed anything.

In mongo:

There should only be 24 records now
dashboard> db.delegationMedia.count()
24

If you have more than 24 records , in mongo drop the collection and redo the steps above:
`db.delegationMedia.drop()`

Should be down to 40mb from 65mb.
image

FYI, I have a UI changes that should drop it by another 12mb...


phile-caci:
Sorry instructions aren't right.. one moment...

kaitlyn-sharo:
This did reduce the number of images in the delegationMedia collection, but in doing so it took those images away from the MRE delegation survey

phile-caci:
Think I'm missing something here, the code that inserts the MRE images doesn't match what I'm seeing in delegationMedia...

for instance, if I query the delegationMedia on prod:
 ``` 
{
    _id: ObjectId('669061a31ed3efdf3b55b621'),
    patientIds: [ 'Patient-N', 'local_soldier_1' ],
    description: 'Male local soldier, broken right shoulder'
},
```

In the update_survey_config.py
```
    tool.add_img_to_db(
        os.path.join(img_dir, "ad_desert_local.png"),
        "**local_soldier_1**",
        "Male, mid-30s; local military uniform.  Tried to break up fight between US soldiers.  Major compound fracture from wall collapse.",
        7,
    )
```

I'll directly resize the encoded images in the db instead...

PR #40, opened 2024-09-17
Itm 582 new probe matcher

kaitlyn-sharo:
This is the new probe matcher for DRE. It takes JSON sim output and matches it to the probes found in the yaml. Once matches are found, probe data is sent to the adept and soartech servers to find alignment data at each target. The matches and alignment data are then stored in the database (humanSimulator collection). The raw data is also stored (humanSimulatorRaw). 

Since there are so many targets, the calls to the server for alignment are slow. To run significantly faster, set RUN_ALIGNMENT on line 9 to False. 

To run, run `python3 dre_probe_matcher.py -i dre_sim_output`. The matches will be outputted to individual json files in the `output` directory (you may want to consider clearing this directory if you already have data in it). You'll see an alignment section and a data section. Focus more on the data section and make sure that the probes and actions match up logically, especially justifications (found in Jungle and Urban adept) and multiple gauze treatments (found in Urban and Desert adept). 

Note that when you run the script, you'll see some warnings: "Could not find match for scene "Probe 9"". This is due to a problem in the sim narrative that we are working out. You will also see one Environment undefined problem. This is an open world scenario. You'll also see that nearly all of the soartech scenarios only have 9 or 10 out of 12 probes found. This is because of problems with the dragging probes. We're looking into this.


phile-caci (approved):
Looks good.

Got the warnings you mentioned and the incomplete number of probes found for SoarTech.

Looked through the data for MJ2, MJ4 and QOL1, the probes and actions seems to match up, I was comparing them to the Review Delegation Material on the UI. 

Looked through the data in mongo.. didn;t look amiss.
`db.humanSimulator.find({"evalNumber":4})`

nextcen-dgemoets:
I must be doing something wrong, but I'm not getting any JSON output from the probe matcher.  The DB is getting updated, but no JSON files are written.  I had cleared out the JSON files before running (based on, "you may want to consider clearing this directory if you already have data in it").

brianpippin:
We will need to make sure the probe matcher works with the new build and the fix to address the one probe from ADEPT that wasn't coming through.  Also you found a work around for detecting dragging correct?

kaitlyn-sharo:
> We will need to make sure the probe matcher works with the new build and the fix to address the one probe from ADEPT that wasn't coming through. Also you found a work around for detecting dragging correct?

I will have to go through the csvs and hand-pick who they were trying to drag for the first participants

dre_probe_matcher.py
```diff
@@ -0,0 +1,649 @@
...
+                for target in targets:
+                    if ('vol' in target and 'vol' not in self.soartech_yaml['id']) or ('qol' in target and 'qol' not in self.soartech_yaml['id']):
+                        continue
+                    st_align[target] = self.get_session_alignment(f'{ST_URL}/api/v1/response', 
+                                            f'{ST_URL}/api/v1/alignment/session?session_id={self.soartech_sid}&target_id={target}',
+                                            match_data,
+                                            self.soartech_sid,
+                                            self.soartech_yaml['id'])
+            except:
+                pass
```
nextcen-dgemoets, line 286: Are you sure we should silently ignore failures to get session alignment from TA1?  If `RUN_ALIGNMENT` is True, won't we want to know if it wasn't obtained?

dre_probe_matcher.py
```diff
@@ -0,0 +1,649 @@
...
+        adept_scenes = self.adept_yaml['scenes']
+        last_action_ind_used = 0
+        match_data = []
+        found = 0
+        total = 0
+        looking_for_justify = False
+        repeating_gauze = False
+        stored_next_id = None
+        # Adept has branching, which makes things a little more difficult
+        first_scene_id = self.adept_yaml['first_scene']
```
nextcen-dgemoets, line 347: Even though ADEPT always uses `first_scene` in DRE, it is not required in the YAML.  It defaults to the first scene in the YAML.

dre_probe_matcher.py
```diff
@@ -0,0 +1,649 @@
...
+                        'ADEPT-DryRun-Ingroup Bias-0.8', 'ADEPT-DryRun-Moral judgement-0.9', 'ADEPT-DryRun-Ingroup Bias-0.9', 'ADEPT-DryRun-Moral judgement-1.0', 'ADEPT-DryRun-Ingroup Bias-1.0']
+                for target in targets:
+                    if ('vol' in target and 'vol' not in self.soartech_yaml['id']) or ('qol' in target and 'qol' not in self.soartech_yaml['id']):
+                        continue
+                    ad_align[target] = self.get_session_alignment(f'{ADEPT_URL}/api/v1/response', 
+                                            f'{ADEPT_URL}/api/v1/alignment/session?session_id={self.adept_sid}&target_id={target}&population=false',
+                                            match_data,
+                                            self.adept_sid,
+                                            self.adept_yaml['id'])
+            except:
```
nextcen-dgemoets, line 583: Same comment as line 286.

nextcen-dgemoets (commented):
Looks good-- major work as usual.  I have a few pretty minor comments/questions in the code.  I scanned through some of the data in the DB and it looks good.  What form does the data in the `humanSimulator` tables appear?  The only reason I ask is that it includes the probe-level `kdma_associations`, which we've been redacting.  Will they appear in the dashboard?

dre_probe_matcher.py
kaitlyn-sharo: Good point, I added a log here

dre_probe_matcher.py
kaitlyn-sharo: You're right. This now will get the first scene, but will default to the first scene in the scenes list if first_scene is not defined

kaitlyn-sharo:
> Looks good-- major work as usual. I have a few pretty minor comments/questions in the code. I scanned through some of the data in the DB and it looks good. What form does the data in the `humanSimulator` tables appear? The only reason I ask is that it includes the probe-level `kdma_associations`, which we've been redacting. Will they appear in the dashboard?

I'm not sure, but that would probably be a dashboard-level problem, not probe-matcher level

nextcen-dgemoets (approved):
@brianpippin   Please see the question above about redacted probe-level kdma_associations, potentially appearing in the dashboard.  I don't know if they will, if they should, and how to avoid it if they shouldn't, but just wanted to flag it.

brianpippin (approved):
This is looking good, as for Darren's question, anything that should be obfuscated will be up to the Dashboard to make sure it doesn't appear, we should make sure that isn't the case and add tickets as necessary when we find that it does.

kaitlyn-sharo:
> This is looking good, as for Darren's question, anything that should be obfuscated will be up to the Dashboard to make sure it doesn't appear, we should make sure that isn't the case and add tickets as necessary when we find that it does.

I did take kdma out of the DRE probe responses in the dashboard play-by-play

PR #41, opened 2024-09-19
remove some probes from adept training scenarios

dereknop:

This script will remove some probes from the adept training scenarios. After running ingest, IO1 should have 8 pages and MJ1 should have 9 pages. Run through the text-based scenario once and verify that we are still getting alignment scores for those two training scenarios by viewing the `alignmentData` field attached to the mongo documents


nextcen-dgemoets (approved):
Both scenarios were shortened to the specified lengths when I completed the text scenarios, and when I reviewed them-- and alignment data was saved when I completed the text scenarios.

brianpippin (approved):
Worked for me.

PR #42, opened 2024-09-20
Optimize getconfig - Reduces patient image sizes

phile-caci:
This code reduces the patient image sizes in delegationMedia. I couldn't get delegation_survey/update_survey_config.py to match the data in prod, patientIds and description are different so just resizing what's in the "url" field (encoded image) in the delegationMedia collection.

I'm also converting PNG to JPEG.  PNG is lossless, JPEG compresses better (I updated the front end to decode for a JPEG instead of a PNG, on my ITM-585 ticket, but that's not in PR yet)

I saved to file the old PNG along with the new JPEG for review, you can delete these files afterwards.

You can check in mongo also
**Old Size:**
db.delegationMedia.dataSize()
30325148 (~30mb)
**New Size:**
db.delegationMedia.dataSize()
2105988 (~2mb)

1. pip install -r requirements.txt
2. . ingest/bin/activate
3. python3 deployment_script.py


brianpippin (approved):
LGTM worked and got all the reduced size outputs.

nextcen-dgemoets:
I ran the script and confirmed the small sizes in the DB, but I can't confirm that smaller sizes are used in the dashboard.  I tried doing a delegation survey (and reviewing one) and opening up some images in another tab, but they seem to be showing as PNGs (URL is `data:image/png;base64,/9j/4AAQSkZJRgABAQAAAQABAAD [etc.]`).

I ran my dashboard using the `ITM-585_UpdateResultsPageforDRE` branch hoping that would contain the "front end to decode for a JPEG instead of a PNG".

Any suggestions to verify the dashboard is using the smaller sizes?  Could it be a caching issue?

nextcen-dgemoets (approved):
Apparently this ticket only addressed MRE image sizes, so there's no way to see the images "live."  But the smaller DB size suggests that everything's fine.

PR #45, opened 2024-09-27
script to append kdma data to text scenarios

kaitlyn-sharo:
We need to display the KDMA values in the dashboard. The most recent commit to probe-matcher will handle that for the sim, but we need the text probes too. Unfortunately, ADEPT's server must have had an update that caused them to lose the data for the session ids we used in week 1 of data collection, so this script not only gets the kdmas, but if the kdma does not exist, it will resend the probes in order to get the kdmas (it will _not_ re-get the alignment). 

You should now be able to see the kdmas if you use https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/125


nextcen-dgemoets:
Does some branch need to be pulled into some other branch?  This PR is showing 43 changed files, including a bunch of pngs.

kaitlyn-sharo:
> Does some branch need to be pulled into some other branch? This PR is showing 43 changed files, including a bunch of pngs.

The probe matcher might need main pulled in

nextcen-dgemoets (approved):
Ran the deployment script which ran the new _0_2_6 script.  In `userScenarioResults`, I now see `kdmas` array in ADEPT scenarios with the appropriate kdma key-value pairs.

I'm told this will appear in the UI in a future PR.

dereknop (approved):

brianpippin (approved):

PR #47, opened 2024-10-01
Probe Matcher Updates

kaitlyn-sharo:
1. Added files from last week's experiments (note that when you run this and look in the dashboard, some data is missing. We are working on getting these local files from OSU. We are currently aware of 107, 104, and 108 missing one or more environments, and 107 has not taken the delegation survey. If you see something other than this, feel free to mention it)
2. Added RUN_ALL option. If RUN_ALL is false, we will look to see if there is already an output file for the input file. If there is and RUN_ALIGNMENT is FALSE, we will not rerun the file. If there is an output file and RUN_ALIGNMENT is true AND the file already has alignment data in it, we will not rerun the file. This should speed up subsequent runs significantly by not rerunning data that isn't changing. And if we do end up having to add something to the output, we can easily switch this variable to True and get the output added to everyone.


dre_probe_matcher.py
```diff
@@ -250,6 +257,27 @@ def __del__(self):
             self.output_adept.close()
 
 
+    def should_file_run(self, filename):
+        '''
+        Looks to see if the input file already has a matching output file. If it does and:
+        1. We are not running alignment OR
+        2. We are running alignment and alignment has already been calculated for this file
```
nextcen-dgemoets, line 264: It looks like this sentence wasn't ended-- as if it needs, "then return False".  But doesn't it also depend on `RUN_ALL`?

nextcen-dgemoets (approved):
Looks good-- I deleted my output files and it ran everything, then I deleted a single output file, ran again, and it only ran that one file.  Results in DB and dashboard look good.

One very minor documentation comment below.

brianpippin (approved):
Had no issues running here.

dre_probe_matcher.py
kaitlyn-sharo: great catch. Fixed!

PR #48, opened 2024-10-01
progress bar

dereknop:
The only change here was adding a progress bar to an old survey version. But to get the images loading in properly for older survey versions I had to re-run some scripts. 

```
cd delegation_survey
python3 update_survey_config.py
enter
y (until it is done)

python3 survey_setup_script.py
y
y
```

That _should_ do it for images properly rendering in [the survey version toggling pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/126)


phile-caci (approved):
Looks  good to me, ran without issue, dashboard delegation survey to 1, and delegation has a progress bar.

for
```
cd delegation_survey
python3 update_survey_config.py
enter
y (until it is done)
```
I used option 1 (I hope that's right since it looked like an edit to surveyConfig1x)

```
cd delegation_survey
python3 update_survey_config.py
enter
1
y (until it is done)
```

nextcen-dgemoets (approved):
I used the same option as @phile-caci  and didn't have any issues.

PR #52, opened 2024-10-03
started adding open world data to the database (no analysis)

kaitlyn-sharo:
Very basic addition of open world data to humanSimulatorRaw

If all data was re-run, openWorld: false would also be added to the other entries. However, the dashboard only looks for openWorld: true, so it is not necessary to rerun all analysis. For now, openWorld will run every time. This just puts the raw data in the database. There is no analysis yet. Please run the dre_probe_matcher to test.


phile-caci (approved):
Looks good,

ran:
`python3 dre_probe_matcher.py -i dre_sim_output`

went from
```
db.humanSimulatorRaw.count({"evalNumber":4,"openWorld": true })
0
```

to:

```
db.humanSimulatorRaw.count({"evalNumber":4,"openWorld": true })
7
```

nextcen-dgemoets (approved):

PR #54, opened 2024-10-04
added latest sim output files

kaitlyn-sharo:
Run python3 dre_probe_matcher.py -i dre_sim_output

You should see only the new files running (because RUN_ALL is false). One (112) will have 10/12 probes for QOL and 11/12 for VOL, another will have 11/12 probes on VOL due to the scenario ending prematurely. Everything else should have full probes. 

Look at the sim dashboard pages to see relevant output. All  scenarios should show up. If a participant's data looks wrong, delete the files for that participant in the "output" directory (where the alignment and matches are stored, NOT the dre-sim-output directory), and run again.


phile-caci (approved):
Looks good to me, no issues in `Human Sim Probe Values` or `Play by Play: Humans in Sim`

PR #55, opened 2024-10-07
Itm 623 data fix

kaitlyn-sharo:
Updated the _0_2_8 and _0_2_9 scripts to calculate alignment comparisons and matches based on most/least aligned 

Before, both scripts were only calculating scores based on adms shown in the delegation survey. Now, _0_2_9 only calculates based on most/least aligned, and _0_2_8 does both, as both values are required

Run the deployment script and check out the admVsTextProbeMatches and humanToADMComparison collections to see adm_types of 'most aligned' and 'least aligned'.

Note that Adept's may not run right now (for _0_2_8) due to a problem in their server

See related dashboard PR https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/138

Additionally, Adept will be changing their compare_session end point to include output separately for MJ and IO, so a separate PR will be needed once they finish that change


nextcen-dgemoets (approved):
The `admVsTextProbeMatches` and `humanToADMComparison` collections now have adm_types of 'most aligned' and 'least aligned.  There were hundreds of errors, but those are ADEPT cases, which are expected.

brianpippin (approved):
Ran as 628 and worked.

PR #56, opened 2024-10-07
Itm 619 compare sim vs text

kaitlyn-sharo:
Get alignment comparisons for the simulator vs text scenarios

run the probe matcher. You'll notice this runs separately from the rest of the code. ADEPT may not run properly right now due to missing session ids in the server.


deployment_script.py
```diff
@@ -10,12 +10,13 @@
...
-        return True 
-    # return true if it is a newer db version
-    return db_version > version_obj['version']
+    return True
+    # collection = mongoDB[VERSION_COLLECTION]
+    # version_obj = collection.find_one()
+    # if version_obj is None:
+    #     return True 
+    # # return true if it is a newer db version
+    # return db_version > version_obj['version']
```
nextcen-dgemoets, line 19: I assume we don't want to merge this change?

dre_probe_matcher.py
```diff
@@ -6,8 +6,9 @@
 from decouple import config 
 
 SEND_TO_MONGO = True # send all raw and calculated data to the mongo db if true
-RUN_ALIGNMENT = True # send data to servers to calculate alignment if true
+RUN_ALIGNMENT = False # send data to servers to calculate alignment if true
```
nextcen-dgemoets, line 9: Is this a deliberate change to the default?  Or just something you wanted to do locally?

nextcen-dgemoets (commented):
> ADEPT may not run properly right now due to missing session ids in the server

Is this what caused dozens of errors like:
- `time data 'Timestamp' does not match format '%m/%d/%Y %I:%M:%S %p'`

deployment_script.py
kaitlyn-sharo: ugh...how many times do I have to make that mistake before I remember!

kaitlyn-sharo:
> > ADEPT may not run properly right now due to missing session ids in the server
> 
> Is this what caused 101 errors like:
> 
> * `time data 'Timestamp' does not match format '%m/%d/%Y %I:%M:%S %p'`

I don't think so...I never got that, I'll look into it

nextcen-dgemoets (approved):
The data was populated in the DB and in the table.

phile-caci (approved):
Ran this after: https://github.com/NextCenturyCorporation/itm-ingest/pull/58

No issues with Adept scenarios

Saw expected? warnings about Probe 9 and 9214/9112/9110 scenarios
Looks good!

brianpippin (approved):
I ran 629, 628, then probe matcher with no issues.

PR #57, opened 2024-10-09
Itm 628 group targets

kaitlyn-sharo:
Script _0_2_9 and _0_3_0 have switched places (the old _0_2_9 became the new _0_3_0). This is because we need to store the text scenario group target alignments before running the adm-text-alignment-match-counter script, since that now includes group target matching scores. 

_0_2_9 appends group targets to _some_ text scenarios (only some participants have been placed in groups).
_0_3_0 has been updated to calculate matches at the group targets.

Run the deployment script to execute both

Then go visit https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/141!


phile-caci (approved):
After running: [ITM-629_RePopAdeptSession](https://github.com/NextCenturyCorporation/itm-ingest/pull/58)
and
https://github.com/NextCenturyCorporation/itm-ingest/pull/57

Looks good, new columns are populated

brianpippin (approved):
I had to rerun 629 right before this, but it ran as expected after I ran it.

phile-caci:
Just FYI, reran this latest update( after 629 ), compared data to Kaitlyn's data.  And we're a match.

PR #61, opened 2024-10-15
Itm 630 open world

kaitlyn-sharo:
Added open world analysis based on the needs of the RQ8 table.

Run the probe matcher (python3 dre_probe_matcher.py -i dre_sim_output). It should only run the open world analyses, due to the global variables set to false. 

Check out the matching dashboard pr (same branch name) and see that table RQ8 is populated.


phile-caci (approved):
LGTM, ran without issue, looked like it was only processing the open world narrative as expected.

dereknop (approved):

nextcen-dgemoets (approved):

PR #63, opened 2024-10-25
script manager

dereknop:

Assuming your database is up to date, `itm_version` should be set to `0.3.1`. This script will look at that current version and determine what scripts need to be run from the `/scripts` to update your database. For the sake of testing these changes, I added two dummy scripts that I will delete before merging this pr. You should see the following print:

Screenshot 2024-10-25 at 9 29 25 AM

After testing, be sure to manually set your `itm_version` back to `0.3.1` since these scripts will be deleted and the next real script will be _0_3_2.

### DIFF
--- a/deployment_script.py
+++ b/deployment_script.py
@@ -1,51 +1,102 @@
 from pymongo import MongoClient
 from decouple import config
-from scripts._0_2_6_add_text_kdmas import get_text_scenario_kdmas
-from scripts._0_2_6_add_text_kdmas import get_text_scenario_kdmas
-from scripts._0_2_8_human_to_adm_comparison import compare_probes
-from scripts._0_2_9_run_group_targets import run_group_targets
-from scripts._0_3_0_percent_matching_probes import find_matching_probe_percentage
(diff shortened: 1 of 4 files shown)


nextcen-dgemoets (approved):
LGTM!  The script output and effect on my DB were what I expected based on the PR description.  Really useful.

phile-caci (approved):
LGTM!
Just a question,  in the past we've had to rerun scripts that were updated. Going forward we should just create a new version of the script then? (Just so we're all on the same page)

dereknop:
> LGTM! Just a question, in the past we've had to rerun scripts that were updated. Going forward we should just create a new version of the script then? (Just so we're all on the same page)

I was unsure how to handle this. For example, probe_matcher and the script that converts yaml to text based config files are kind of special cases since they are iterated on. My objective with this was to zero in on the one-time use scripts. For example, if you miss an ingest pr you don't need to hunt down what you missed because this will just look back and run the ones you missed.

dereknop:
I don't know if making a whole new file for those special cases is the best approach. I am open to ideas on this though because I wasn't sure how to account for those. Using regex and the naming convention in `/scripts` was much easier to account for.

PR #65, opened 2024-10-29
Open world order fix

kaitlyn-sharo:
Some patients were showing order = 9 or 10, but should only be 1-8. Duplicate patients were showing up in the engagement list. This fixes that to make sure when we get the order, there is only one of each patient. Rerun the probe matcher (it will be fast if you have already run it on all these files, because it will only redo the open world files)


nextcen-dgemoets (approved):
After running the script, there was the same number of rows, but none of the order columns had 9 or 10.

PR #68, opened 2024-11-01
script for fixing 20240913

dereknop:

pid 20240913 was entered twice for two different participants. This script groups them based on their start time so they can be distinguished from each other in the dashboard. After running the script, instead of 10 documents in `userScenarioResults` with pid `20240913`, there should be 5 with `20240913A` and a start time of `Tue Oct 01 2024 15:11:25 GMT-0400 (Eastern Daylight Time)` and 5 documents with `20240913B` and a start time of `Tue Oct 01 2024 15:34:46 GMT-0400 (Eastern Daylight Time)`. You should be able to view this change reflected on the participant view of http://localhost:3000/text-based-results for IO2, MJ1, and MJ2 scenarios

### DIFF
--- a/scripts/_0_3_3_fix_202409113.py
+++ b/scripts/_0_3_3_fix_202409113.py
@@ -1,0 +1,28 @@
+def fix_pid(mongo_db):
+    text_scenario_results = mongo_db['userScenarioResults']
+
+    documents = list(text_scenario_results.find(
+        {"participantID": "202409113"}
+    ).sort("startTime", 1))  
+    
+    unique_times = sorted(set(doc["startTime"] for doc in documents))
+    if len(unique_times) != 2:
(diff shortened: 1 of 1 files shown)


kaitlyn-sharo (commented):
This works as intended, but I wonder if we should add 113A and 113B to the participantLog so that the other tables also update? Now as it stands, none of these results appear in any of the other tables because they don't show anything that's not in the participantLog

dereknop:
> This works as intended, but I wonder if we should add 113A and 113B to the participantLog so that the other tables also update? Now as it stands, none of these results appear in any of the other tables because they don't show anything that's not in the participantLog

Good point, I did not think of this. I can edit this same script to update the participate log. Will do after I get the kdma dashboard pr up

kaitlyn-sharo (approved):
Works for me

nextcen-dgemoets (approved):

PR #70, opened 2024-11-11
Itm 670

dereknop:
[Dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/159)

```
cd text_based_scenarios
python3 convert_yaml_to_json_config.py
python3 add_patient_pictures_p1.py
```

This should add the new json configs for the text-based scenarios to collection `textBasedConfig` and the new soartech images to `textBasedImages`


phile-caci (approved):
LGTM, looks like it show up in the UI and is working correctly.

Question:  Regarding the phase1 scenarios, should we be using the ones Darren posted in Teams this morn? The ones here, are a little different.

BEFORE:
```
dashboard> db.textBasedConfig.count()
27
dashboard> db.textBasedImages.count()
83
```

AFTER:
```
dashboard> db.textBasedConfig.count()
38
dashboard> db.textBasedImages.count()
155
```

dereknop:
> Question: Regarding the phase1 scenarios, should we be using the ones Darren posted in Teams this morn? The ones here, are a little different.

It depends... the soartech ones should be an exact match. So if there are discrepancies there, I should fix that. As far as Adept goes, their two training scenarios have been narrowed down to shorten the time it takes and MJ2, MJ4, and MJ5 all have small changes in some of the action mappings to make it more compatible with the format of the text based scenarios. I worked these changes out with the Adept team leading up to DRE.

nextcen-dgemoets (approved):
Database was updated as described.

PR #72, opened 2024-11-18
Itm 704 ingest eval scenarios

phile-caci:
Ingested Eval 5 Eval Scenarios

`python deployment_script.py`

Will updated db version to 0.3.4 and add 9 scenarios to the scenarios table.

`db.scenarios.find({"evalNumber":5})`
`db.scenarios.count({"evalNumber":5})`

### DIFF
--- a/phase1/scenarios/phase1-adept-eval-MJ2.yaml
+++ b/phase1/scenarios/phase1-adept-eval-MJ2.yaml
@@ -1,0 +1,1899 @@
+id: DryRunEval-MJ2-eval
+# v3.1.0
+#   - Added rapport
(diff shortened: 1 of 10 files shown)


dereknop (approved):
Worked for me after I remembered ingest scripts work a whole lot better when your mongo container is actually running... 
Screenshot 2024-11-18 at 9 41 15 AM

nextcen-dgemoets (approved):
After running deploy, the DB shows 9 scenarios with `evalNumber` of 5.
The scenarios have `kdma_associations` that aren't all set to zero.

PR #74, opened 2024-11-20
Itm 708 ph1 probe matcher

kaitlyn-sharo:
Not many changes here:
1. Moved the old probe matcher files somewhere else so they didn't get in the way of the new one
2. Based ph1 probe matcher off of dre
3. Switched probe names for ST based off new yamls
4. Changed yaml access location to the correct directory
5. Changed adept endpoint for alignment to new endpoint
6. Changed evalNumber and name to 5/Phase 1 Eval
7. Changed file names to be ph1
8. Updated soartech targets 

To test, grab some of the old adept sim outputs and some of your new ST outputs (from testing the phase1 simulator) and put them in a directory. Then run the new probe matcher (python3 ph1_probe_matcher.py -i [input-directory]). The global variables should be right for now, as we don't have anything to run comparisons on yet. Check the `output` directory and look through the alignment scores and probe matches. Data should also appear in humanSimulator and humanSimulatorRaw (in mongo)


nextcen-dgemoets (approved):
Looks great.  After running the script, I saw the expected records in the `humanSimulator` and `humanSimulatorRaw` tables, and they matched the JSON files in my `output` directory.

The only slightly surprising thing I noticed is that the results included alignment for group targets.  I guess these will get updated after we do group targets in January?

kaitlyn-sharo:
> Looks great. After running the script, I saw the expected records in the `humanSimulator` and `humanSimulatorRaw` tables, and they matched the JSON files in my `output` directory.
> 
> The only slightly surprising thing I noticed is that the results included alignment for group targets. I guess these will get updated after we do group targets in January?

I noticed that too. It's coming straight from the adept endpoint, so I figured it was okay.

nextcen-dgemoets:
> I noticed that too. It's coming straight from the adept endpoint, so I figured it was okay.

True-- I guess I meant SoarTech where we're explicitly adding it to a list of alignment targets.  That said, if it's already in ADEPT, there's not much reason to exclude it from ST.  We'll just have some "junk" data in the DB that will be overwritten later.

dereknop (approved):

brianpippin (approved):
I didn't see any issues.

phile-caci (approved):
No issues on my end.. got files from Darren to ingest... LGTM!

PR #78, opened 2024-11-22
remove page titles

dereknop:
```
cd text_based_scenarios
python3 convert_yaml_to_json_config.py
```

Should remove page titles on pages. Can use review-text-scenario to test.

Before:
Screenshot 2024-11-22 at 10 34 00 AM

After:
Screenshot 2024-11-22 at 10 32 52 AM

Note: I believe the grab for the configs is cached so you may have to do a hard refresh or close the tab and go back.

### DIFF
--- a/text_based_scenarios/convert_yaml_to_json_config.py
+++ b/text_based_scenarios/convert_yaml_to_json_config.py
@@ -159,6 +159,7 @@ def partition_doc(scenario, filename, eval_type):
 
         page = {
             'name': scene['id'],
+            'title': ' ',
             'scenario_id': scenario_id,
             'scenario_name': scenario['name'],
             'elements': []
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
The titles are gone.

kaitlyn-sharo (approved):
It worked BUT it would be nicer if it wasn't a space (just an empty string instead) because now I can highlight a tiny space at the top of the text scenarios.

dereknop:
> It worked BUT it would be nicer if it wasn't a space (just an empty string instead) because now I can highlight a tiny space at the top of the text scenarios.

I know it is annoying... survey js automatically used the `name` field if the title is empty though and I use that `name` field in a couple other places.

phile-caci (approved):
Titles are empty for me!

PR #80, opened 2024-12-02
added phase 1 sim files to repo

kaitlyn-sharo:
Updated the repo to include phase 1 sim files in directory ph1_sim_files (avoiding the word "output" this time @nextcen-dgemoets)

I also added logic to the ph1 probe matcher to automatically remove those files/directories that are from testing (i.e. do not have a valid pid) and those that are from dre (prior to 11/24).

There was also a missing probe match in MJ2 (probably from a minor yaml change), so I updated that.

To run: `python3 ph1_probe_matcher -i ph1_sim_files`

A dashboard PR will be up later today to update some fields that are missing data, but you can check the dashboard for basic information, and check mongo for expected entries


dereknop (approved):

phile-caci (approved):
LGTM!

PR #82, opened 2024-12-04
Itm 741 ph1 comparisons

kaitlyn-sharo:
Added a new script ph1_weekly.py for the jenkins job that will run the comparison scripts. 

Probe Matcher: comparison/alignment portions have now been updated. A 'weekly' option has been added, which sets the global variables so that the only actions are those that are required (new files analyzed fully, old files without comparisons/alignments will be compared/aligned, as needed).

Script 0_2_8: updated to work for eval 4 and eval 5

Script 0_3_0: updated to work for eval 4 and eval 5

To test, just run ph1_weekly.py. You should see RQ5 Alignment scores populate, as well as the Del|ADM alignment score in RQ1/3/4


ph1_weekly.py
```diff
@@ -0,0 +1,21 @@
...
+from scripts._0_3_0_percent_matching_probes import find_matching_probe_percentage
+
+
+if __name__ == '__main__':
+    # set up mongo
+    client = MongoClient(config('MONGO_URL'))
+    db = client.dashboard
+
+    # run 'weekly' version of probe matcher
+    os.system('python3 ph1_probe_matcher.py -i ph1_sim_files -w')
```
nextcen-dgemoets, line 14: For me, my system python 3.x is simply `python`, so I needed to update the script.  But that's no problem-- as long as it works on prod.

nextcen-dgemoets (approved):
RQ5 Alignment scores populated, and there were values for `Alignment score (Delegator|Observed_ADM (target))` in RQ1/3/4.

ph1_weekly.py
phile-caci: This ok on my end, I have python3 and python sym linked to python 3.9 on prod.

phile-caci (commented):
Ran for me without issue, however, for it to run on PROD can we change these (I normally just change these myself on PROD before running ingest):

in `ph1_probe_matcher.py`
```
ADEPT_URL = "https://darpaitm.caci.com/adept/" #config("ADEPT_URL")
ST_URL = "https://darpaitm.caci.com/soartech/" #config("ST_URL")
```
to
```
ADEPT_URL = config("ADEPT_URL")
ST_URL = config("ST_URL")
```

Also I default the .env to those URL's above, so you "shouldn't" have to change it for local development.

I can also make the change you don't care, thanks!

brianpippin:
If we make the change Phi suggested, then it is good to go on my end.

PR #83, opened 2024-12-05
Deployment script fix

dereknop:
There was an error where the deployment script was not calling the first function in a given script, instead it was calling the first one by alphabetical order. Now it should look for `def main`. I included a test script where there is one function that both appears before `main` and is ahead of it alphabetically. You should see running `python deployment_script.py` that it hits the right function. 

After testing, set your db version back to what it was previously. I will delete the test script before merging.


phile-caci (approved):
Works for me!

nextcen-dgemoets:
Should this PR go back and restore the use of `config()` when it would otherwise have been used?

nextcen-dgemoets:
For that matter, it could go back and change other script's first functions to `main()`.  I guess it's unlikely anybody will need to run really old ingest scripts, as we usually just pass around an updated database.

Another thought is that maybe the "Your script MUST have 'def main'" comment should be in the README where it talks about running a new script.

kaitlyn-sharo:
> For that matter, it could go back and change other script's first functions to `main()`. I guess it's unlikely anybody will need to run really old ingest scripts, as we usually just pass around an updated database.
> 
> Another thought is that maybe the "Your script MUST have 'def main'" comment should be in the README where it talks about running a new script.

I agree, I feel like with this change we should go back through all the old scripts to make them compatible with this.

dereknop:
Fair point, I'll go back and change them to main so if need be we can call them

dereknop:
Update: I am going to merge this in so it doesn't cause issues with any upcoming ingest pr but will create a ticket for myself to go back through all of these because I need to do something else rn

PR #85, opened 2024-12-10
update old scripts

dereknop:
I updated the old scripts to conform to the new convention expected by the deployment script. The easiest way to test is to set your `itm_version` back a couple in Mongo Compass and then run `python3 deployment_script.py`. Ensure the deployment script detects the proper scripts (e.g. if your version is 0.3.6, it runs 0.3.7 and 0.3.8) and runs them without error.

### DIFF
--- a/scripts/_0_1_7_add_eval_4_scenarios.py
+++ b/scripts/_0_1_7_add_eval_4_scenarios.py
@@ -13,12 +13,13 @@ def load_json_file(folder: str, file_name: str) -> dict:
             encoding='utf-8-sig') as json_file:
         return json.loads(json_file.read())
     
-def main(ta1_folder, mongoDB):
-    ta1files = [f for f in os.listdir(ta1_folder)]
+def main(mongoDB):
+    path = "dre/scenarios/"
+    ta1files = [f for f in os.listdir(path)]
     ta1files.sort()
 
     for file in ta1files:
-        file_name = os.path.join(ta1_folder, file)
+        file_name = os.path.join(path, file)
         with open(file_name) as f: 
             yaml_obj = yaml.safe_load(f)
             yaml_obj["evalNumber"] = 4
--- a/scripts/_0_1_7_add_eval_4_scenarios.py
+++ b/scripts/_0_1_7_add_eval_4_scenarios.py
@@ -13,7 +13,7 @@ def load_json_file(folder: str, file_name: str) -> dict:
             encoding='utf-8-sig') as json_file:
         return json.loads(json_file.read())
     
-def load_ta1_yaml_files(ta1_folder, mongoDB):
+def main(ta1_folder, mongoDB):
     ta1files = [f for f in os.listdir(ta1_folder)]
     ta1files.sort()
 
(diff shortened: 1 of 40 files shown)


phile-caci (approved):
LGTM, I set the db to 0.3.5 and the depoy script ran 3.6, 3.7 and 3.8.

scripts/_0_1_7_add_eval_4_scenarios.py
```diff
@@ -13,7 +13,7 @@ def load_json_file(folder: str, file_name: str) -> dict:
             encoding='utf-8-sig') as json_file:
         return json.loads(json_file.read())
     
-def load_ta1_yaml_files(ta1_folder, mongoDB):
+def main(ta1_folder, mongoDB):
```
nextcen-dgemoets, line 16: I'm not sure how far back we have to go, but is this going to work?

scripts/_0_2_8_human_to_adm_comparison.py
```diff
@@ -29,7 +29,7 @@
     "DryRunEval-MJ5-eval": ['Probe 1', 'Probe 1-A.1', 'Probe 1-B.1', 'Probe 2', 'Response 2-A.1-B', 'Response 2-B.1-B', 'Response 2-B.1-B-gauze-u', 'Response 2-A.1-B-gauze-sp', 'Probe 2-A.1-A.1', 'Probe 2-B.1-A.1', 'Probe 2-A.1-B.1-A.1', 'Probe 2-B.1-B.1-A.1', 'Probe 3', 'Probe 4']
 }
 
-def compare_probes(mongoDB, EVAL_NUMBER=4):
+def main(mongoDB, EVAL_NUMBER=4):
```
nextcen-dgemoets, line 32: This will always use an `EVAL_NUMBER` of 4, no?

scripts/_0_3_0_percent_matching_probes.py
```diff
@@ -1,6 +1,6 @@
 import utils.db_utils as db_utils
 
-def find_matching_probe_percentage(mongoDB, EVAL_NUMBER=4):
+def main(mongoDB, EVAL_NUMBER=4):
```
nextcen-dgemoets, line 3: This will always use an `EVAL_NUMBER` of 4, no?

kaitlyn-sharo: No! It will not. I have a script that is supposed to run weekly (it's called something like weekly ph1 script) that uses this and another script for phase 1, so I converted them to work for 4 and 5

scripts/_0_2_8_human_to_adm_comparison.py
kaitlyn-sharo: This is the other script I was mentioning in my other comment

kaitlyn-sharo: @dereknop I haven't looked at this yet, but please make sure if you haven't that the weekly script is updated with these new names

nextcen-dgemoets:
I think this PR could back and restore the use of `config()` where it was removed because of the prior issue.  Specifically, `_0_3_6_rerun_soartech_sessions.py` and `_0_3_8_update_dre_kdmas.py`.

nextcen-dgemoets (approved):
LGTM.  All of the scripts should work with the updated deployment script.

kaitlyn-sharo (approved):
Seems good to me!

PR #88, opened 2024-12-18
added script to fix some database inaccuracies

kaitlyn-sharo:
327 took 336's survey (336 did not take the survey), so the pid needed to be changed as well as the admChoiceProcess for any incorrect targets loaded.

316 had 9 text results. We removed the later duplicates

Run the deployment script and see in the progress table that 336 no longer has a survey, 327 does, and 316 has 5 instead of 9 text responses

### DIFF
--- a/scripts/_0_4_0_db_cleanup.py
+++ b/scripts/_0_4_0_db_cleanup.py
@@ -1,0 +1,30 @@
+from bson.objectid import ObjectId
+
+'''
+Update some database entries to match reality and remove duplicate entries
+'''
+
+def main(mongo_db):
+    survey_collection = mongo_db['surveyResults']
+    # 327 took 336's survey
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (commented):
Is there anything we should do about the case where PID="undefined" and where the PID was, e.g., 66c7d3c15993ceaa74816309?

That said, all of the changes you identified are seen in the Progress Table:
itm-775

kaitlyn-sharo:
> Is there anything we should do about the case where PID="undefined" and where the PID was, e.g., 66c7d3c15993ceaa74816309?
> 

I don't know about the undefined pid, but the prolific id ones, from what I understand, are only on text scenarios (not surveys) that were already completed once all the way through with a valid PID. So they will just be ignored

nextcen-dgemoets (approved):

jaudick (approved):

PR #89, opened 2024-12-18
Itm 772 approve users

kaitlyn-sharo:
Run the deployment script to set all current users as "approved". To be used in tandem with the matching 772 dashboard PR


phile-caci (approved):
No issues, all users are not set to approved: true

nextcen-dgemoets (approved):
After running the script, all 71 users in my DB had `approved=True`.

PR #90, opened 2024-12-19
ITM-769:  Update SoarTech Phase 1 KDMA scores to add scalar value

nextcen-dgemoets:
Adapted the [DRE version of this script](https://github.com/NextCenturyCorporation/itm-ingest/blob/main/scripts/_0_3_8_update_dre_kdmas.py) for Phase 1, accounted for some DB changes, and added the first round of manual adds for incomplete sim sessions.

To test, run the deployment script and check RQ8 table to see that the "Participant KDMA" column of the table has values in it for SoarTech (use the TA1 filter).  And/or check your db to see that the values are there.

You can run this against the production AWS SoarTech server, but it will be slowwww.  If you can run the server locally, pull the latest from the `phase1` branch, put the unredacted scenarios in `ta1_server/data/scenarios/current/`, and change your ingest `.env` to point to your local server.

Some things you should expect, as of this writing:
- You will see two "No profile for <PID>" warnings, one of which will add, "Updating with manual data."
- You will see two PIDs of "undefined" when processing SoarTech Text KDMAs.
- You will see eight weird-looking PIDs like "6635054dfb8caedd407ca84b" which are duplicates/errors.
- There are only 33 rows of sim output (not including warnings).  More will be added as the data comes in.

### DIFF
--- a/scripts/_0_4_2_update_ph1_kdmas.py
+++ b/scripts/_0_4_2_update_ph1_kdmas.py
@@ -1,0 +1,224 @@
+import requests
+from decouple import config 
+ST_URL = config("ST_URL")
+UPDATE_DATABASE = True # Update the mongo database tables for text, sim, and adm
+VERBOSE_OUTPUT = True # If True, displays scalar KDMA values for each entry
+DISPLAY_CSV = False # If True, writes csv-like output for text and sim kdma data
+txt_and_sim_data = {}
+
+"""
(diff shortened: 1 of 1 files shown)


kaitlyn-sharo (approved):
worked for me!

brianpippin (approved):
Worked and showed for me, lets get this on production.

PR #92, opened 2024-12-23
improved script 028 to have 'run-new-only' mode

kaitlyn-sharo:
Script _0_2_8 (which runs as part of the phase 1 weekly script) is VERY slow. Nobody likes it. This adds a new mode "run-new-only" that will only run the comparisons that do not already appear in the database. The weekly script runs this fast version. 

If your database is up to date, when you run this (by running the weekly script), you should see the errors from the surveys that are missing, and a few others that are expected, but it should run quickly. If you remove the "True" parameter, it should revert to its old, slow behavior.


phile-caci (approved):
No issues running on my end.

nextcen-dgemoets (approved):
Worked for me and was much faster.  I now like Script _0_2_8 and it can play in our reindeer games again.

PR #93, opened 2024-12-23
base probe matcher runs on database, not just output directory

kaitlyn-sharo:
Perhaps this has only been a problem for me, but I will sometimes run the probe matcher locally, then download the most recent database which has not had probe matcher run. It will undo all of my database probe matcher changes, but since I have the files locally, probe matcher does not rerun those files and repopulate the database.

This makes sure probe matcher runs based on the database as well as the output directory, not just the local directory. To test, remove an entry from your local database and rerun probe matcher with the weekly setting. You should see it stop _once_ to recalculate the probes and alignment for that one entry. 

As a more important, related change:
The KDMAs added from Darren's 0_4_2 script were being overwritten because of the run_comparison function. This ensures that when the weekly script is run, the alignment data for comparison is grabbed from the database rather than from local files (the local files don't have kdmas). To test, run Darren's script and the old probe matcher (with the weekly setting on). For a faster run, only run the sim part of Darren's script. Notice in the dashboard (check http://localhost:3000/humanSimParticipant) that all the sim kdmas disappear. Now do it again, this time running this updated probe matcher. All the data should remain.

### DIFF
--- a/ph1_probe_matcher.py
+++ b/ph1_probe_matcher.py
@@ -246,16 +246,29 @@ class ProbeMatcher:
         then return False in order to skip the analysis of this file.
         '''
         run_this_file = True
-        if not RUN_ALL and os.path.exists(filename):
+        mongo_id = None
+        if 'adept' in self.environment:
+            mongo_id = self.participantId + '_ad_' + self.environment.split('.yaml')[0]
+        elif 'qol' in self.environment or 'vol' in self.environment:
+            mongo_id = self.participantId + '_st_' + self.environment.split('.yaml')[0]
+        else:
+            mongo_id = self.participantId + '_dre_open_world'
+        found = list(mongo_collection_matches.find({'_id': mongo_id}))
+        if not RUN_ALL and ((SEND_TO_MONGO and len(found) > 0 and os.path.exists(filename)) or (not SEND_TO_MONGO and os.path.exists(filename))):
             if RERUN_ADEPT_SESSIONS and 'adept' in self.environment:
                 return run_this_file
             if not RUN_ALIGNMENT:
                 run_this_file = False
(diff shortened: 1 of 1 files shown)


phile-caci (approved):
Not seeing any issues on my end.

I removed 202411302 (sim)
On main the values for 302 didn't repopulate with ph1_weekly.
On this branch, running the ph1_weekly doesn't remove any of the sim values and repopulated 302.

ph1_probe_matcher.py
```diff
@@ -246,16 +246,29 @@ def should_file_run(self, filename):
...
             if RERUN_ADEPT_SESSIONS and 'adept' in self.environment:
                 return run_this_file
             if not RUN_ALIGNMENT:
                 run_this_file = False
             if RUN_ALIGNMENT:
-                f = open(filename, 'r', encoding='utf-8')
-                data = json.load(f)
-                if len(list(data.get('alignment', {}).keys())) > 1 and 'sid' in data.get('alignment', {}):
-                    run_this_file = False
+                # TODO: update this based on database instead of output directory
```
nextcen-dgemoets, line 263: Is this still needed?  Isn't this what you just added?

nextcen-dgemoets (approved):
Not seeing any issues on my end.
I removed all 202411337 entries from `humanSimulator` (with eval 5).
On `main` only the `Desert Open World Narrative` values for 337 repopulated with `ph1_weekly`.
On this branch, running `ph1_weekly` doesn't remove any of the sim values and repopulated 337.

ph1_probe_matcher.py
kaitlyn-sharo: oops - note to self from one day to the next. removed

PR #94, opened 2024-12-27
Calculate dre comparisons using ph1 server

kaitlyn-sharo:
Jennifer wanted to see the text/adm comparisons (based on the delegation survey only (i.e. not most/least aligned)) for the DRE participants using the Phase 1 server, since ADEPT changed their comparison calculation. 

This required several steps:
1. Re-send text probes from all 3 ADEPT text scenarios (MJ1, IO1, and the eval scenario) to the phase 1 server to create a new session
2. Create the "mini-adms" (aka delegation adms, which use only a subset of probes depending on IO vs MJ) on the phase 1 server
3. Generate the comparison for those two newly-create session and store it 

The text scenario now has an optional parameter ph1SessionId which corresponds to the session id on the phase 1 server without compromising the dre session id. The comparison collection has an optional parameter ph1_server which shows whether the score was gotten from ph1 or dre. 

While working on this task, I discovered that at some point, the entire admMedic collection was wiped clean, so all of our eval4 medics are gone. Because of the way naming works, I had to create a new script to repopulate the admMedic collection in the correct order. First, it runs eval 3, then eval 4, then eval 5, but for eval 5 the naming starts from scratch. 

To run:
`cd delegation_survey`
`python3 rerun_all_delegation.py`
`cd ..`

Note that this script is 0_4_5. 0_4_4 is the group targets, which will likely be merged before this one. If you test out of order,  you'll have to play around with the script numbers. If this gets merged first, I will change it to 044.

View https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/221 afterwards

### DIFF
--- a/delegation_survey/ph1_dre_adms_for_delegation.py
+++ b/delegation_survey/ph1_dre_adms_for_delegation.py
@@ -1,4 +1,4 @@
-import sys
+import sys, argparse
 sys.path.insert(0, '..')
 from pymongo import MongoClient
 from decouple import config 
@@ -12,8 +12,8 @@ Pushes each individual adm to the admMedics collection in mongo.
 '''
 LOGGER = Logger('ADM Converter')
 UPDATE_MONGO = True
-DELETE_MONGO = True
-EVAL = 5
+DELETE_MONGO = False
+EVAL = 4
 
 # reserve 'O' for omnibus
 names = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z']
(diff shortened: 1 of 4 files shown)


delegation_survey/ph1_dre_adms_for_delegation.py
```diff
@@ -12,8 +12,8 @@
 '''
 LOGGER = Logger('ADM Converter')
 UPDATE_MONGO = True
-DELETE_MONGO = True
-EVAL = 5
+DELETE_MONGO = False
+EVAL = 4
```
nextcen-dgemoets, line 16: Just want to make sure that this is deliberately the new default.

nextcen-dgemoets (approved):
It would be nice to have some logging to show incremental progress in such a long-running script, but everything looks good.  One small sanity check question below.

delegation_survey/ph1_dre_adms_for_delegation.py
kaitlyn-sharo: Yes, this was deliberate. This shouldn't have to run for its original purpose anymore because we have already generated the survey, but it definitely needs to run with the new script and these defaults help with that fix

phile-caci (approved):
LGTM!  No issues running. humanToADMComparison looks good.

PR #95, opened 2025-01-02
added script for repopulating adept servers

kaitlyn-sharo:
To run, just run the deployment script. This should update the session ids for adms and text scenarios, including within the comparison and mini-adms collections. No new comparisons or alignments are run

Please consider testing locally, if possible


scripts/_0_4_6_adept_repop_ph1.py
```diff
@@ -0,0 +1,148 @@
...
+            for k in entry:
+                if isinstance(entry[k], dict) and 'questions' in entry[k]:
+                    if 'probe ' + k in entry[k]['questions'] and 'response' in entry[k]['questions']['probe ' + k] and 'question_mapping' in entry[k]['questions']['probe ' + k]:
+                        response = entry[k]['questions']['probe ' + k]['response'].replace('.', '')
+                        mapping = entry[k]['questions']['probe ' + k]['question_mapping']
+                        if response in mapping:
+                            probes.append({'probe': {'choice': mapping[response]['choice'], 'probe_id': mapping[response]['probe_id']}})
+                        else:
+                            print('could not find response in mapping!', response, list(mapping.keys()))
+            print("TXT PROBES: " + str(probes))
```
nextcen-dgemoets, line 65: I think we can remove this logging?  Also the ADM equivalent, line 93.
(And yes, I know I was asking to add logging earlier.)

nextcen-dgemoets (approved):
Looked good, ran fine.

The ADEPT session IDs were updated in RQ1, RQ2.2, RQ3, and RQ4, while the SoarTech session IDs were unchanged.  I ran against a local ADEPT server instance.

scripts/_0_4_6_adept_repop_ph1.py
kaitlyn-sharo: I built off of the dre script, which had this. Since you had asked for logging, I went against my better judgement and kept it haha. I removed the probe printing now, but not the session ids, as it might be helpful for testing

nextcen-dgemoets: The probe logging was only for an edge case where we need to log logging the full probe if `VERBOSE_LOGGING` was true

dereknop (approved):

PR #96, opened 2025-01-02
RQ1/3/4 data table fix 

dereknop:
http://localhost:3000/research-results/rq1

Problem pids
Screenshot 2025-01-02 at 2 32 32 PM

These pids are all missing an entry in the column 'Alignment score (Delegator|Observed_ADM (target))' for the misaligned row.

Easier to do a spot check using the following filters 
TA2: kitware, TA1: adept, Attributes: Moral,  Scenarios: AD-1

Run `python deployment_script.py` and then `python ph1_weekly.py`

Those columns should now be populated


kaitlyn-sharo (approved):

nextcen-dgemoets:
It looks like you need to remove _0_4_6.

dereknop:
I thought I did but evidently I did not, thanks

dereknop:
202411369 was closed out of by the participant before the results from the sever attached to the document. I added in a script that re runs this document and gets the kdma, mostLeastAligned

kaitlyn-sharo:
Weekly script now contains eval 4 as well, for a missing value due to the same reason as the missing eval 5 values

kaitlyn-sharo (approved):

nextcen-dgemoets (approved):
Before I ran the script, the specified IDs had empty `Alignment score (Delegator|Observed_ADM (target))` values, but after the script (and after I got a fresh DB), the values were populated.
202411369 also has more data populated in various tables as well.

PR #97, opened 2025-01-02
Itm 785 group targets

kaitlyn-sharo:
Adds phase 1 group targets. Just run the deployment script. Check table 2.1 for results.

This only adds ST group targets as we are working on some Adept issues currently.


brianpippin (approved):
This ran for me and now my RQ 2.1 is populated.

dereknop (approved):
ran with no errors, 2.1 populated with st group targets

nextcen-dgemoets:
Now that I updated my DB to last night's snapshot, I'm seeing the correct data.

PR #98, opened 2025-01-02
Comparison fix

kaitlyn-sharo:
Adept added a new secret endpoint for comparisons. This makes sure the current code gets updated with that new endpoint. 

The only two files that used comparison are the probe matchers (only phase 1 needed the update) and 028.

028 needs to be compatible with the old and new evals. Phase 1 probe matcher only needs the new version. 

I have updated the weekly script for the probe matcher to rerun comparison. This is based on old code that always reruns all of 028, so we're good there.

Please check to make sure these endpoint conversions are correct. 

To test, you should just be able to run the weekly script. Then check table RQ1 in the delegator vs adm target column on prod vs dev to make sure those values change, and check table RQ6 for the sim data updates


kaitlyn-sharo:
When this is run on prod:
Run the weekly script
Update the database itm version to 0.4.3
Run the deployment script
Ideally, just run compare_probes(db, 4, True) and compare_probes(db, 5, True) (the middle two commands in the weekly script)

This will:
1. Rerun the probe matcher to refresh sim/text and sim/adm comparisons using the new endpoint
2. Rerun the adm/text comparisons (most/least and misaligned/aligned/baseline)
3. Fix the database version since we had accidentally skipped a number
4. Rerun the DRE comparisons against the Phase 1 server using the new endpoint (updates the misaligned/aligned/baseline)
5. Update the missing values in the RQ1 table due to weird adm output
6. Rerun only the missing values to update their comparisons

kaitlyn-sharo:
Update: run `py1_one_time.py`

dereknop (approved):
Ran without error and I can see updated results when comparing to prod

brianpippin (approved):

nextcen-dgemoets (approved):
Okay finally.

Yes, looks good:  `Alignment score (Delegator|Observed_ADM (target))` was updated in RQ1 (and the non-DRE rows of RQ4).  And RQ6 data are now properly associated with ADEPT unlike in prod, where you have things like SoarTech -> MJ.

PR #100, opened 2025-01-06
added update to repopulation script to look for conditional probes

kaitlyn-sharo:
MJ2 and MJ4 have conditional probes hidden within other probe sections. Therefore, the old method of searching for one probe per probe section in the db does not work. This adds the search for that conditional probe. To check, run the attached script against your local adept server (you can put it in the top level of ingest)
[compare_alignments.py.zip](https://github.com/user-attachments/files/18322649/compare_alignments.py.zip)

Check differences.csv and see that there are 14 matches for everything that has a valid PID and is not MJ5. MJ5 is a work in progress

### DIFF
--- a/adept_repop_ph1.py
+++ b/adept_repop_ph1.py
@@ -4,7 +4,7 @@ from scripts._0_3_0_percent_matching_probes import main as find_matching_probe_p
 
 
 MONGO_URL = config('MONGO_URL')
-ADEPT_URL = config("ADEPT_URL")
+ADEPT_URL = "http://127.0.0.1:8080/" #config("ADEPT_URL")
 
 
 SCENARIO_MAP = {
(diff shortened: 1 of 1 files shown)


ph1_probe_matcher.py
```diff
@@ -17,7 +17,7 @@
 EVAL_NAME = 'Phase 1 Evaluation'
 
 
-ADEPT_URL = config("ADEPT_URL")
+ADEPT_URL = "http://localhost:8080/" #config("ADEPT_URL")
```
nextcen-dgemoets: Did you mean to check in this change to ensure we use our local server?

nextcen-dgemoets (approved):
After the latest push, results look good running with local ADEPT and prod SoarTech.
Just left with what expected logging, including not finding ADMs for group target IDs.

dereknop (approved):

PR #102, opened 2025-01-09
Run ph1 data through dre server

kaitlyn-sharo:
Run the deployment script.

If you want to run locally (which I highly recommend, so we do not pollute the prod adept servers, and because it's much faster), you will have to:
1. Make a second clone of the adept server repo
2. Switch to tag 3.9.3 (git checkout 3.9.3)
3. Remove the redacted files and replace them with the phase 1 files
4. Run your original local adept server on 8080
5. Run this new repo on 8087
6. Update the .env accordingly

This script generates two csvs:
The ph1_server_old_comparison_vs_new.csv file is the dre data and ph1 data sent to the ph1 server. Then the new comparison endpoint was called and the old comparison endpoint on the ph1 server was called and both values are recorded.
 
The dre_server_ph1_data.csv file is all of the phase 1 data sent to the dre server, and the values requested are side-by-side with the phase 1 data values from the phase 1 server. 

There are several steps in this script, all should be pretty clear from the comments throughout. Please check the logic and endpoints to make sure they match what you would expect.

I also added a parameter to 028 so that we do not have to sit through ST rerunning every time.

### DIFF
--- a/scripts/_0_4_9_adept_experiment.py
+++ b/scripts/_0_4_9_adept_experiment.py
@@ -1,0 +1,338 @@
+from decouple import config 
+import requests, csv
+from scripts._0_4_4_adept_human_adm_compare_DRE_PH1_Server import main as dre_to_ph1
+from adept_repop_ph1 import main as repop_ph1, SCENARIO_MAP
+import utils.db_utils as db_utils
+from scripts._0_2_8_human_to_adm_comparison import main as compare_probes
+
+PH1_URL = config("ADEPT_URL")
+DRE_URL = config("ADEPT_DRE_URL")
+
+def main(db):
+    refill_PH1_sessions = 'http:' in PH1_URL # if ph1 server is local, refill sessions
+
+    comparisons = db['humanToADMComparison']
+
+    comparison_file = open('ph1_server_old_comparison_vs_new.csv', 'w', encoding='utf-8')
+    comparison_writer = csv.writer(comparison_file)
+    comparison_writer.writerow(['PID', 'Eval_Number', 'Text_Session', 'ADM_Session', 'Text_Scenario', 'ADM_Scenario', 'ADM_Author', 'ADM_Type', 'ADM_Alignment_Target', 'Old_Endpoint_Score', 'New_Endpoint_Score'])
+
+    # all PH1 sessions should be in PH1 server
+    if refill_PH1_sessions:
+        print("Repopulating Phase 1 server with Phase 1 data")
+        repop_ph1(db)
+        compare_probes(db, 5, False, True)
+    # Run old comparison endpoint on PH1 sessions in PH1 server
+    ph1_comparisons = comparisons.find({'evalNumber': 5, 'dre_server': {'$exists': False}, 'adm_scenario': {'$exists': True}})
+    for comp in ph1_comparisons:
+        res = requests.get(f'{PH1_URL}api/v1/alignment/compare_sessions?session_id_1={comp['text_session_id']}&session_id_2={comp['adm_session_id']}').json()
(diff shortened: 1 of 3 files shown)


scripts/_0_4_9_adept_experiment.py
```diff
@@ -0,0 +1,316 @@
...
+    for x in probes:
+        requests.post(f'{server}api/v1/response', json={
+            "response": {
+                "choice": x['choice'],
+                "justification": x["justification"],
+                "probe_id": x['probe_id'],
+                "scenario_id": x['scenario_id'],
+            },
+            "session_id": adept_sid
+        })
```
nextcen-dgemoets, line 147: Should this use the util send_probes?  Hmm, although I didn't use it in `_0_4_2`, so maybe there's a reason not to.

scripts/_0_4_9_adept_experiment.py
```diff
@@ -0,0 +1,316 @@
...
+                page['dreTxtAlignment' if EVAL_NUMBER == 5 else 'ph1TxtAlignment'] = txt_alignment
+
+        # store data in survey pages since we gather data based on what was seen in the survey
+        survey_collection.update_one({'_id': res['_id']}, {'$set': res})
+        print(f'updated survey data for {pid}')
+
+
+def get_target_score(EVAL_NUMBER, url, session_id, target):
+    '''
+    Endpoint chnaged between DRE and PH1 - DRE server for eval 5 can get just the target.
```
nextcen-dgemoets, line 112: Typo: `chnaged`

scripts/_0_4_9_adept_experiment.py
```diff
@@ -0,0 +1,316 @@
...
+        # go through each page to find the data we need
+        for pageName in res['results']:
+            if 'Medic' in pageName and ' vs ' not in pageName:
+                page = res['results'][pageName]
+                target = page['admTarget']
+                # only look for adept pages
+                if 'ADEPT' not in target:
+                    continue
+
+                # find comparison values from dre server from database
```
nextcen-dgemoets, line 59: Looks like this comment should be more generic as code is no longer DRE-specific.

scripts/_0_4_9_adept_experiment.py
```diff
@@ -0,0 +1,316 @@
...
+                comparison_params = {('dre_server' if EVAL_NUMBER == 5 else 'ph1_server'): True, 'evalNumber': EVAL_NUMBER, 'adm_type': page['admAlignment'], 'pid': pid, 'adm_author': page['admAuthor'], 'adm_scenario': page['scenarioIndex']}
+                comparison = comparison_collection.find_one(comparison_params)
+                if not comparison:
+                    print('could not find comparison document with parameters', comparison_params)
```
nextcen-dgemoets, line 63: I got about 10 of these per server.  Why does this happen?  Not a PR review comment, just a question.

nextcen-dgemoets (approved):
Really nice work.  Minor questions below and in the other ticket.

scripts/_0_4_9_adept_experiment.py
kaitlyn-sharo: heh - you keep correcting my typos and see if I include comments next time!! :)

scripts/_0_4_9_adept_experiment.py
kaitlyn-sharo: I'm pretty sure we send the adm probes in a different format than text. So text probes are {'probe': {probe data}}, whereas adm probes are {probe_data}. You can see in db_utils we have x['probe']['choice'], whereas here we just have x['choice']. This is an opportunity for future improvement, but not here.

dereknop (approved):

PR #104, opened 2025-02-06
More info easily available in test collection 

dereknop:
This is a quality-of-life improvement for the `test` collection that contains our adm runs. Before, we had to click through the history array to see the adm name, scenario, and target. Now all of that info should be available at the root level, making it easier to query and sift through. 

Screenshot 2025-02-06 at 9 52 06 AM


kaitlyn-sharo (approved):
super helpful!

nextcen-dgemoets (approved):
Love it!

PR #106, opened 2025-02-11
Itm 852 jan eval

kaitlyn-sharo:
Run `python3 jan_one_time.py`. It will run the probe matcher for the january eval sim files as well as the comparison script (that compares adm to human alignment) and the matcher script (that finds what percentage of adm and human probes match). Then check out the dashboard PR.

Please make sure your db is updated before running.


nextcen-dgemoets (approved):
Other than having to change the command line to `python`, it ran (very slowly) with no errors.

kaitlyn-sharo:
> Everything ran. Curious for these if it would be better to just have this be a script in the usual script folder that gets run once and calls probe matcher instead of having a different paradigm that requires going onto the server and running a specific command. Just wondering if there is a reason for not putting it into scripts, since those are only suppose to run once as well.

When I did this for phase 1, I think I did it this way because I was having trouble with imports, but I can give it a try if it's easier

brianpippin:
> > Everything ran. Curious for these if it would be better to just have this be a script in the usual script folder that gets run once and calls probe matcher instead of having a different paradigm that requires going onto the server and running a specific command. Just wondering if there is a reason for not putting it into scripts, since those are only suppose to run once as well.
> 
> When I did this for phase 1, I think I did it this way because I was having trouble with imports, but I can give it a try if it's easier

Let's not worry about this one now since everyone tested it like this, let us see if we can do that for the next one.

kaitlyn-sharo:
> > > Everything ran. Curious for these if it would be better to just have this be a script in the usual script folder that gets run once and calls probe matcher instead of having a different paradigm that requires going onto the server and running a specific command. Just wondering if there is a reason for not putting it into scripts, since those are only suppose to run once as well.
> > 
> > 
> > When I did this for phase 1, I think I did it this way because I was having trouble with imports, but I can give it a try if it's easier
> 
> Let's not worry about this one now since everyone tested it like this, let us see if we can do that for the next one.

will do

PR #107, opened 2025-02-18
run all adms against all humans

kaitlyn-sharo:
To run, run the deployment script. There are three new scripts:

The first (053) sends the january eval data to the DRE server. 

The second (054) sends all ADEPT phase 1 ADMs to the DRE server (not subsets of probes - full probe responses)

The third (055) runs every single participant's text responses from DRE, PH1, and JAN against every single ADM with the same scenario. This uses the ADEPT DRE comparison endpoint every time, while ST changes based on DRE vs PH1/JAN. It also runs the percent-probe-match script (what percent of the time did the ADM and human choose the same probe response). 

The special thing about this, aside from it being every single combination of target/adm/human per scenario, is that we are not taking a subset of probes this time. 

This takes forever to run, because every text scenario gets compared against all 4 adms across all the targets for that scenario (which could be anywhere between 8-16, or something like that. Group targets are included). It would be difficult to run locally (as in, with local TA1 servers) because you'd need to run the proper other scripts first to populate your local database with all the correct session ids, but it is doable. Feel free to do this if you'd like. If not, be prepared to wait a full business day for the script to finish. 

A csv (full_probe_comparrisons.csv) is outputted with the data from the third script.

NOTE: Right now, the third script is only running Eval 5 (PH1), per Jennifer's request

There is a possibility your mongo cursor will time out during this process. I have tried to put precautions in place so this will not happen, but it doesn't seem to always work. If it times out, run it again - the script skips everything that has already been run and will continue from where it left off

### DIFF
--- a/scripts/_0_5_3_jan_eval_to_dre_server.py
+++ b/scripts/_0_5_3_jan_eval_to_dre_server.py
@@ -3,7 +3,6 @@ import requests
 from adept_repop_ph1 import SCENARIO_MAP
 import utils.db_utils as db_utils
 
-PH1_URL = config("ADEPT_URL")
 DRE_URL = config("ADEPT_DRE_URL")
 
 def main(db):
--- a/scripts/_0_5_3_jan_eval_to_dre_server.py
+++ b/scripts/_0_5_3_jan_eval_to_dre_server.py
@@ -1,0 +1,76 @@
+from decouple import config 
+import requests
+from adept_repop_ph1 import SCENARIO_MAP
+import utils.db_utils as db_utils
+
+PH1_URL = config("ADEPT_URL")
+DRE_URL = config("ADEPT_DRE_URL")
+
+def main(db):
+    # send all JAN sessions to DRE server
+    print("Populating DRE server with JAN Eval data")
+    send_JAN_to_DRE(db)
+
+def send_JAN_to_DRE(mongoDB, EVAL_NUMBER=6):
+    '''
+    Populates the DRE server with PH1 data
+    '''
+    text_scenario_collection = mongoDB['userScenarioResults']
+
+    data_to_use = text_scenario_collection.find(
+        {"evalNumber": EVAL_NUMBER}
+    )
+
+    # start by creating sessions in dre server for phase 1 data
+    sessions_by_pid = {}
+    for entry in data_to_use:
+        scenario_id = entry.get('scenario_id')
+        dre_id = entry.get('dreSessionId', None)
+        data_id = entry.get('_id')
+        pid = entry.get('participantID')
+        
+        new_id = None
+        if 'adept' in scenario_id and dre_id is None:
+            scenario_id = SCENARIO_MAP[scenario_id]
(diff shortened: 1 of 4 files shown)


dereknop (approved):

nextcen-dgemoets (commented):
This finally finished (overnight).  It exited at one point yesterday evening after about 100 records with a mongo cursor timeout, but when I re-ran it, it continued and eventually completed.

A few observations on the output, which all might be normal/expected:
- There are 2792 rows with `-1` in the `Alignment score (Participant|target)` column.  These are all group targets, and include every scenario except MJ5.  In other words, only MJ5 produced actual scores in this column when selecting a group target, if I'm not mistaken.
- Sometimes in the numeric fields (alignments and match), the value is a string, other times it's a number.  This is probably an artifact of opening the .csv in MS Excel, but I thought I'd mention it.
- This output includes 641 rows with the bad PIDs (e.g. `670e2a711df0a045aea8de04`) and another 161 rows with a PID of `undefined`.  I assume this should just be included in the data and ignored?  Not related to this ticket, but is there an argument to remove this from the DB?

NOTE:  I did not run this with this morning's commit: "updated code to work for evals 4 and 6 properly"

kaitlyn-sharo:
> This finally finished (overnight). It exited at one point yesterday evening after about 100 records with a mongo cursor timeout, but when I re-ran it, it continued and eventually completed.
> 
> A few observations on the output, which all might be normal/expected:
> 
> * There are 2792 rows with `-1` in the `Alignment score (Participant|target)` column.  These are all group targets, and include every scenario except MJ5.  In other words, only MJ5 produced actual scores in this column when selecting a group target, if I'm not mistaken.
> * Sometimes in the numeric fields (alignments and match), the value is a string, other times it's a number.  This is probably an artifact of opening the .csv in MS Excel, but I thought I'd mention it.
> * This output includes 641 rows with the bad PIDs (e.g. `670e2a711df0a045aea8de04`) and another 161 rows with a PID of `undefined`.  I assume this should just be included in the data and ignored?  Not related to this ticket, but is there an argument to remove this from the DB?
> 
> NOTE: I did not run this with this morning's commit: "updated code to work for evals 4 and 6 properly"

Thanks for pointing out those missing group targets. I think that is a bug and I think I know how to fix it.

I have now added float casts to all the scores just to make sure they are all numbers.

I have removed all entries that are not in the participant log.

The commit from this morning does not affect the functionality, except fixing a few things for 4 and 6 so that if Jennifer ever wants them we can run them easily.

nextcen-dgemoets (approved):
LGTM.  Compared to the previous run:
- The old list had 16065 rows.  Without the undefined and sessionId-style PIDs, it goes down to 15265.  The new table only has 13105 rows, but (I'm told) that's because fixing the group targets involved removing some rows that shouldn't have been there, while fixing others so that they have a value.
- All of the -1 alignment scores are gone-- yay!
- All scenarios are represented in the group target rows.
- All of the numbers look like numbers (not text).

PR #108, opened 2025-02-18
calculate jan eval data using dre server

kaitlyn-sharo:
Run the deployment script. This script is similar to 049_adept_experiment, which ran all the dre data through the phase 1 server and the phase 1 data through the dre server, if you're looking for a reference.

The script sends all delegation adms that participants encountered during the JAN Eval experiment to the DRE server. It then calculates the necessary comparisons using the DRE server.

To see the results, checkout the corresponding dashboard PR

You may want to comment out 055 for now, so it doesn't have to run before reviewing this

### DIFF
--- a/scripts/_0_5_6_adept_human_adm_compare_JAN_DRE_Server.py
+++ b/scripts/_0_5_6_adept_human_adm_compare_JAN_DRE_Server.py
@@ -1,0 +1,234 @@
+from decouple import config 
+import requests
+import utils.db_utils as db_utils
+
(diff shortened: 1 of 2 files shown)


dereknop:
Definitely comment out 055

dereknop:
I ran 053 and 054, skipped 055 and ran this script. Got a timeout. Giving it another go

dereknop (approved):
second time worked fine, I think it maybe timed out first time because my computer went into sleep mode

nextcen-dgemoets (approved):

PR #111, opened 2025-03-03
Itm 874

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-874)

Reads in the manually compiled spreadsheet of any error in delegation materials seen in the jan eval and excludes problem probes from delegator to adm comparison. Script should take a few minutes to run. It should rescore using both dre and ph1 adept server to update both corresponding columns in rq1/3/4 tables. I have a lot of prints to help track the processing, ultimately, I expect 130 documents to be updated (65 p1 and 65 dre server comparisons)

`python deployment_script.py`

Note: This is not final because I need to incorporate Kaitlyn's 0_5_8 script into this as well.


scripts/_0_5_9_fix_jan_comparisons.py
```diff
@@ -0,0 +1,370 @@
...
+                            score_change = new_score - old_score
+                            score_changes.append({
+                                'pid': pid,
+                                'scenario': doc_scenario,
+                                'old_score': old_score,
+                                'new_score': new_score,
+                                'change': score_change,
+                                'percent_change': (score_change / old_score) * 100 if old_score != 0 else 0,
+                                'is_dre': is_dre
+                            })
```
kaitlyn-sharo, line 342: Curious if any of this is needed. I didn't see it mentioned in the ticket request. Was it just for debugging?

scripts/_0_5_9_fix_jan_comparisons.py
```diff
@@ -0,0 +1,370 @@
...
+
+#regex stuff to handle sometimes 0.7 vs 07 in the targets in mongo collections
+def normalize_target(target):
+    if re.search(r'-\d{2}$', target):
+        return re.sub(r'-(\d)(\d)$', r'-\1.\2', target)
+    elif re.search(r'-\d\.\d$', target):
+        return re.sub(r'-(\d)\.(\d)$', r'-\1\2', target)
+    return target
+
+def mini_adm_run_fixed(mongo_db, pid, scenario, adm_name, target, problem_probes, original_doc):
```
kaitlyn-sharo, line 38: Looks like adm_name is not used

scripts/_0_5_9_fix_jan_comparisons.py
```diff
@@ -0,0 +1,370 @@
...
+        'alignment_target': {'$in': [original_target, alt_target]}
+    }
+    
+    matching_adms = list(adm_collection.find(query))
+    
+    if not matching_adms:
+        print(f"    No matching ADM found for PID {pid}, scenario: {test_scenario}, targets: [{original_target}, {alt_target}]")
+        return None
+    
+    adm = matching_adms[0]
```
kaitlyn-sharo, line 60: I'm not sure I agree with this. There are 4 ADMs that could have the same scenario/target/evalNumber. This is probably where adm_name was meant to go. In this case, it looks like you're only getting the first matching adm, which will be the same adm alignment/author every time, even when it's supposed to be maybe the 4th. 

I would recommend adding adm name to the query and then making sure there's only one found (using something like count_documents, or just get the length of the list in line 54), otherwise you may be editing the wrong data, but maybe I'm just missing something. There's a lot to keep track of here

kaitlyn-sharo: Note: I did print out the length of matching_adms, and it is consistently coming up as 4.

scripts/_0_5_9_fix_jan_comparisons.py
```diff
@@ -0,0 +1,370 @@
...
+        print(f"      {i+1}. {probe}")
+
+   
+    adept_sid = requests.post(f'{api_url}api/v1/new_session').text.replace('"', "").strip()
+    print(f"    Created new ADM session: {adept_sid}")
+    
+    # send proebs
+    for i, x in enumerate(probe_responses):
+        print(f"      Sending probe {i+1}/{len(probe_responses)}: {x['probe_id']} (choice: {x['choice']})")
+        response = requests.post(f'{api_url}api/v1/response', json={
```
kaitlyn-sharo, line 115: no need for "response" here as it is never used

kaitlyn-sharo (changes requested):
There's a lot of printing going on here. But it all happens so fast and creates so many prints that if I wanted to, I couldn't really go back to investigate because terminal runs out of space to hold it all. If it's important to see this, could we move it to a text file instead?

If it's not important, can we remove it? It would make the code a lot easier to look through and understand and cut down on a ton of lines. Also, maybe consider using \t instead of spacing at the beginning of the prints for easier readability

Some print statements and comments have typos in them (i.e. proebs and probelm), but I'm not that particular about that. Just wanted to mention it.

The rest of the code made sense to me, though I would have recommended using the already-built mini_adm_run for some of the code to remove redundancy. It would also mean that rerunning would not create new session ids in the adept db because we'd have the updated sessions stored in our own db.

scripts/_0_5_9_fix_jan_comparisons.py
dereknop: Not needed besides for debugging prints will remove

scripts/_0_5_9_fix_jan_comparisons.py
dereknop: Thanks, yea I think I just spaced on adding the adm name to the query like you said. Will change

kaitlyn-sharo (approved):
I think this looks good now.

PR #112, opened 2025-03-05
Added 'eval' argument to probe matcher

kaitlyn-sharo:
Download the latest database backup from mongo. Look at the progress table, filter by Jan Eval, and see that there are some clearly not Jan Eval participants listed (i.e. pids are not 700-level)

This happened over the weekend when the weekly script ran. To fix it, I added an 'eval' argument to the probe matcher so that when the weekly script runs, it runs with eval 5, and when the jan_one_time script runs, it runs with eval 6.

Rerun `python3 ph1_weekly.py`

Notice that all of the data is back where it should be. You can also check mongo to ensure no extra/duplicate entries are being put in the database.

When this is merged, we will have to rerun the weekly script from jenkins

The ph1_weekly script has several functions commented out since we have had to do so much work recalculating scores. I don't want to risk this messing with that.

### DIFF
--- a/ph1_weekly.py
+++ b/ph1_weekly.py
@@ -15,8 +15,8 @@ if __name__ == '__main__':
 
     # run script that compares humans to adms (RQ1/3/4 column Alignment score (Del|ADM))
     # RQ 5 Alignment score (Participant|ADM (most, least))
-    compare_probes(db, 4, True)
-    compare_probes(db, 5, True)
+    # compare_probes(db, 4, True)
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
Using last night's archive, I could see the spurious "Jan 2025" pids in the progress table.  After running the ph1_weekly script, those pids were no longer in the table.

dereknop (approved):

PR #114, opened 2025-03-13
added development script to get csv of kdmas

kaitlyn-sharo:
New directory: dev_scripts - this is for scripts that do not need to be run on production and may need to be run more than once. For example, this `get_text_kdmas` script outputs a csv and does nothing to the database. Grabbing the latest database and running it should be sufficient for its purpose. 

The script gets all the text scenarios from evals 4, 5, and 6, filters by ADEPT and valid PIDs (pids that appear in the participant log), and gets all kdmas available: combined kdma value (overall), narrative (MJ2, 4, 5), and non-narrative (MJ1, IO1). It puts all of this into a csv called "text_kdmas"

Note that you will get a few warnings about individual_kdma missing for 733. 733 and 734 had a lot of extra entries. Not sure what we're doing about this yet. 733 still has several entries in the outputted csv, so ignore the warning for now.

To run, just run `python3 dev_scripts/get_text_kdmas.py` and check the output csv

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ output
 */jen-work
 adept-evals/eval4/*.json
 adept-evals/eval4/*.py
+/*.csv
+dev_scripts/*.csv
--- a/dev_scripts/get_text_kdmas.py
+++ b/dev_scripts/get_text_kdmas.py
@@ -1,0 +1,89 @@
+import csv
+from pymongo import MongoClient
+from decouple import config 
+
+IO = 'Ingroup Bias'
+MJ = 'Moral judgement'
+
+def get_text_kdmas(mongo_db, output='text_kdmas.csv'):
+    '''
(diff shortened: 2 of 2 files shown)


nextcen-dgemoets:
Should this remove duplicates-- that is, if the MJ and IO values are the same?
e.g., the output includes:
```
202501728,6,narr,0.4359466201578393,0.3025598748343483
202501734,6,narr,0.4359466201578393,0.3025598748343483
```
If not, how would these ADM runs be distinguished?  (not a question for this ticket so much, but for ITM-893)

nextcen-dgemoets (approved):

dereknop (approved):
I love the idea of the new folder moving forward. Script appeared to work as intended with 499 rows populated in my CSV. Checked out the code and it all made sense to me.

PR #115, opened 2025-03-17
ITM-870

dereknop:
Creates the collection that is needed to toggle styles via the admin dashboard page. 

### DIFF
--- a/scripts/_0_6_1_add_style_collection.py
+++ b/scripts/_0_6_1_add_style_collection.py
@@ -1,0 +1,3 @@
+def main(mongo_db):
+    ui_collection = mongo_db['uiStyle']
+    ui_collection.insert_one({'version': 'updated'})


nextcen-dgemoets (approved):
The `uiStyle` collection was added when I ran the script.  Amazing work.

brianpippin (approved):
Approving, but for future ref, this PR might be superfluous because Mongo will auto create a collection that doesn't exist when you first add something to it as a heads up.

dereknop:
> Approving, but for future ref, this PR might be superfluous because Mongo will auto create a collection that doesn't exist when you first add something to it as a heads up.

Understood, the only reason I made this was that the dashboard side pulls down the value to determine styling and I thought this may be preferable to leaving logic on the dashboard side that accounts for no document being found. I definitely will not make it a habit of making ingest PRs this trivial.

PR #116, opened 2025-03-20
Added script for dre alignment on ph1 adept adms

kaitlyn-sharo:
Task: Add script that will run each ADM/Target for the ADEPT Scenarios against the DRE endpoint, and add a field to the record so that both the P1E alignment score is there, with the additional DRE alignment score

We had already run each ADM for the ADEPT scenarios (ph1) against the DRE endpoint in 054. We had stored the dresessionids, so just needed to grab those and get the alignment using the DRE endpoint. 

The alignment is stored next to the ph1 alignment in the last history index. Check mongo to see that dre_alignment appears for these adms.

Working on a dashboard PR now to visualize this. Draft until it's done

Run the deployment script. You should not get any errors. (Note: this will not get the alignment for group targets)

### DIFF
--- a/scripts/_0_6_1_dre_adm_alignment_adept.py
+++ b/scripts/_0_6_1_dre_adm_alignment_adept.py
@@ -16,12 +16,19 @@ def main(mongoDB):
         history = adm['history']
         dre_session_id = history[-1]['parameters'].get('dreSessionId')
         target = adm["alignment_target"]
+        if "-Group-" in target:
+            continue
         if dre_session_id is None:
             print(f'Error getting dre session id from {adm["adm_name"]} - {adm["scenario"]} - {target}')
             continue
         
         # dre version of alignment score
         alignment = requests.get(f'{DRE_URL}api/v1/alignment/session?session_id={dre_session_id}&target_id={target}&population=false').json()
+
+        if 'score' not in alignment:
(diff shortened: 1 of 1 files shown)


scripts/_0_6_1_dre_adm_alignment_adept.py
```diff
@@ -0,0 +1,38 @@
+# _0_5_4 sends all phase 1 adms (ADEPT only) to the DRE server
+# this gets the alignment using those session ids and stores it in the db
```
nextcen-dgemoets, line 2: This comment should be updated.

nextcen-dgemoets (commented):
I had to rename the file to 062 because of [ITM-870](https://github.com/NextCenturyCorporation/itm-ingest/pull/115).

But the script ran fine:
itm-903

scripts/_0_6_1_dre_adm_alignment_adept.py
kaitlyn-sharo: To what? I wanted to note what the script did. Did you want me to just delete it?

nextcen-dgemoets: Oh sorry, I misunderstood.  I thought _0_5_4 was this script, but re-reading it I see my mistake.

dereknop (approved):
Ran with no issues

nextcen-dgemoets (approved):

PR #118, opened 2025-03-25
KITWARE EXPERIMENT: script to generate adm groups with new comparisons

kaitlyn-sharo:
Check the [corresponding dashboard PR](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/269) to see results.

Run the deployment script. 
- The newest script (064) grabs the eval 7 adms, gets sets of 3 where the adm and target are the same, but the scenario is different. 
- Grab the alignment target from those adms and find the participant text entry (or entries) that match with that target. 
- Send the adm probes to the dre adept server and store that session id. 
- If the text entry kdmas that match were from narr or training only,send those specific probes to the server and stores the session id. Otherwise grab the dreSessionId that was already stored. 
- Send those two session ids to the dre server and store
- All values are stored in a new multiKdmaData database so that the processing is fully kept out of the dashboard

### DIFF
--- a/scripts/_0_6_4_setup_p2e1.py
+++ b/scripts/_0_6_4_setup_p2e1.py
@@ -61,10 +61,10 @@ def main(mongo_db):
             group_count += 1
         adm_groups[adm_name][target_id].append(adm)
 
+    completed_groups = 0
     for adm_name in adm_groups:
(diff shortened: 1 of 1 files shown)


dereknop (approved):

nextcen-dgemoets (approved):
LGTM.

PR #119, opened 2025-03-26
Itm 915

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-915)

`python deployment_script.py`

Will do the following:
- add to `scenarios` collection the appropriate yaml files for eval number 7. These are a duplicate of the p1 files but we have queries in the dashboard that pull the scenarios based on evalNumber so this is to maintain that functionality. 
- goes through adm runs and pulls out more information abt them to the top level. This is re using an older script I ran in the past but on the new data
- removes the test data that leaked into phase 1

### DIFF
--- a/multi-kdma/scenarios/phase1-adept-eval-MJ5.yaml
+++ b/multi-kdma/scenarios/phase1-adept-eval-MJ5.yaml
@@ -1,0 +1,1954 @@
+id: DryRunEval-MJ5-eval
+# v1.8.1
+#  - Changed injuries that were discoverable to visible.  With the restricted actions, there wasn't a way to actually discover the injury.  And this way it matches the unstructured text.
+# v1.8.0
+#  - Added rapport
+# v1.7.2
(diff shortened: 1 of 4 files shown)


multi-kdma/scenarios/phase1-adept-eval-MJ5.yaml
kaitlyn-sharo, line 1: Just curious if we really want to duplicate these files in the repo or if we can just pull from the phase 1 directory?

nextcen-dgemoets: The duplication struck me as well, but I was going to let it slide.  However, if there's a better solution, that would be great.

kaitlyn-sharo (commented):
Seemed to work. Happy to approve once we either remove the duplicates or have a good reason for them

kaitlyn-sharo (approved):

PR #120, opened 2025-03-31
fixed bug in 064 to get kdmas

kaitlyn-sharo:
At some point, I erroneously removed the line that grabs the kdmas from each adm in order to fill up the multiKdmaData database. This adds that line back in.

To run, get the newest db, then revert your db version to 0.6.3 and run the deployment script. It should run much faster because now all of the session ids are stored, so it will just recalculate alignment. 

Check out the table on the dashboard (RQ2, Phase II experiment 1, 2.3) before and after. Before, you'll see lots of repeating kdma values. After, you should see a variety.


nextcen-dgemoets (approved):
Before:
repeat

After:
after

dereknop (approved):
Getting unique scores now. My first few rows look like Darren's. Awesome we didn't have to wait for the whole script to run through fresh again.

PR #121, opened 2025-04-01
Itm 926 update kitware

kaitlyn-sharo:
Run the deployment script (no db change needed - took Brian's advice from yesterday and added 0_6_6 to call 0_6_5). 

This removes the no-duplicate-target rule and adds the pids to the database. It will take a few minutes to run because there are some humans who had not been run through this script due to the no-duplicate-target rule, so their sessions need to be run and stored. 

Notice that the RQ 2.3 Phase II experiment table in the dashboard has increased in row number (1151 to 1303). You can check this by filtering and looking at the "out of ____ rows" number.


nextcen-dgemoets (approved):
The data looks good and I have 1303 rows.

dereknop (approved):

PR #123, opened 2025-04-04
Itm 929

dereknop:
Accounts for baseline runs. Pulls information out of history to the top level to be used in gql queries, removes the uuid from the adm name strings. 


dereknop:
> Looks good-- the uuids are removed and baseline adms show up in 2.3. The only thing that looked odd is that shouldn't the ADM Type column for the baseline ADMs be set to `baseline`?

Yea that is a one-line thing on the dashboard side. I would have put it on the dev to main had I known about the new baseline runs. I'll open up the world's smallest pr for that right now.

nextcen-dgemoets (approved):

PR #124, opened 2025-04-04
added distance endpoint data to the db

kaitlyn-sharo:
Adds delegator|target alignment values using the P1E server's distance-based endpoint to the text collection (see distance_based_most_least_aligned)

Adds delegator|adm comparison alignment values using the P1E server's distance-based endpoint to the humanToADMComparison collection (see distance_based_score)

Adds adm alignment values using the P1E servers' distance-based endpoint to the admTargetRuns collection (see history[-1][response] distance_based_score)

Check out the corresponding dashboard PR to see the results in RQ 1/3/4, but also check out the database to see that only evals 5 and 6 are updated with this data.

### DIFF
--- a/scripts/_0_6_8_adept_p1e_old_endpoints.py
+++ b/scripts/_0_6_8_adept_p1e_old_endpoints.py
@@ -92,7 +92,7 @@ def main(mongo_db):
         sys.stdout.flush()
         if 'distance_based_score' in adm['history'][-1]['response']:
             continue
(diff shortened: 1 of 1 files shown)


dereknop (approved):
Ran with no issues. Spot checked in mongo compass and saw all of the additions you mentioned

nextcen-dgemoets (approved):
Too many tickets!  Didn't mean to withhold my approval here, just lost track of it.

PR #125, opened 2025-04-07
added baseline multi-kdma analysis

kaitlyn-sharo:
Two updates:
1. Updated 065 to use the Phase 1 server with distance-based endpoints instead of the DRE server
2. Added 069 to add baseline data to the multiKdmaData collection

Run the deployment script. Script 069 runs two scripts at the beginning:
- 067, which updates the names of the ADMs to be cleaner and also calls 051 to move adm metadata to the top level
- 065, the updated version. This needs to be rerun in order to enact the DRE -> PH1 change

After all of that is updated, we grab one baseline ADM per scenario (3 total, since all runs against a scenario should be the same). We run distance-based alignment from the PH1 server on every participant against the ADM.

Check out the dashboard to see results.

Due to the DRE -> PH1 change, this takes a little while to run, as we have to create and store more session ids in the database

### DIFF
--- a/scripts/_0_6_5_setup_p2e1.py
+++ b/scripts/_0_6_5_setup_p2e1.py
@@ -1,9 +1,8 @@
 from decouple import config 
 import requests, os, csv, sys
 import utils.db_utils as db_utils
-from scripts._0_5_3_jan_eval_to_dre_server import main as update_dre_server
 
-ADEPT_DRE_URL = config("ADEPT_DRE_URL")
+ADEPT_URL = config("ADEPT_URL")
 
 SCENARIO_MAP = {
     'DryRunEval-MJ2-eval': 'AD1',
@@ -25,8 +24,7 @@ def main(mongo_db):
     Also calculates the comparison alignment between the human who generated the target 
     and the adm run against that "synthetic" target
     '''
-    # run the jan_eval_to_dre_server script to make sure all jan eval text sessions have been sent to the dre server
-    update_dre_server(mongo_db)
+
     # run the dev script to get text. store in list for easy indexing
     os.system('python3 dev_scripts/get_text_kdmas.py')
     f = open('text_kdmas.csv', 'r', encoding='utf-8')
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (approved):
I think this branch needs the `f"` fix, but other than that everything ran well.

kaitlyn-sharo:
> I think this branch needs the `f"` fix, but other than that everything ran well.

I merged that in locally, but forgot to push! Thanks for the reminder

dereknop:
```
_0_6_8_adept_p1e_old_endpoints.py", line 95
    alignment = requests.get(f'{SERVER_URL}api/v1/alignment/session?session_id={adm['history'][-1]['parameters']['session_id']}&target_id={adm['alignment_target']}&population=false').json()
                                                                                     ^^^^^^^
SyntaxError: f-string: unmatched '['
```

I think since we have `f'` it doesn't like the single quotes for accessing history. 
This worked for me
`alignment = requests.get(f'{SERVER_URL}api/v1/alignment/session?session_id={adm["history"][-1]["parameters"]["session_id"]}&target_id={adm["alignment_target"]}&population=false').json()`

kaitlyn-sharo:
> ```
> _0_6_8_adept_p1e_old_endpoints.py", line 95
>     alignment = requests.get(f'{SERVER_URL}api/v1/alignment/session?session_id={adm['history'][-1]['parameters']['session_id']}&target_id={adm['alignment_target']}&population=false').json()
>                                                                                      ^^^^^^^
> SyntaxError: f-string: unmatched '['
> ```
> 
> I think since we have `f'` it doesn't like the single quotes for accessing history. This worked for me `alignment = requests.get(f'{SERVER_URL}api/v1/alignment/session?session_id={adm["history"][-1]["parameters"]["session_id"]}&target_id={adm["alignment_target"]}&population=false').json()`

This should have already been fixed, did you pull?

dereknop:
Thought I did during standup maybe it was before your earlier comment idk. Either way, my mistake running again on a fresh db to make sure I'm getting the same stuff

dereknop (approved):

PR #126, opened 2025-04-09
Itm 923

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-923)

The `redo.py` script should take in a command line argument and run that script regardless of whatever the current version of your local db is. For example, `python redo.py 067` should run `_0_6_7_cleanup_adm_results.py`. I uploaded a dummy script `_0_0_0_hi_from_test_script.py` that does nothing but print so we can test easily. 

`python redo.py 000`

Test that you can run that as many times as your heart desires. I have created a Jenkins job that takes a parameter `script_number` and runs `redo.py`. Once this gets merged, I can test that job with the 000 dummy script (I don't want to test with a script that could potentially mess anything up with the prod db).


nextcen-dgemoets (approved):
Love this little script.  I ran it as many times as my heart desired.

nextcen-dgemoets:
One thought, though-- when the script fails, could it show the traceback?

dereknop:
> One thought, though-- when the script fails, could it show the traceback?

Should now should full traceback instead of just error message

kaitlyn-sharo (commented):
Two things:
1. I wonder if we should put this in dev_scripts (maybe along with adept_repop, jan_one_time, ph1_one_time, and maybe everything that isn't deployment_script? Probably needs a little more discussion). Not necessary though
2. If I do something crazy like redo.py 0067, it runs 006, which would not be my expected behavior. Maybe we should make sure there are only 3 numbers in that parameter and error otherwise. Also, when we get to bigger script numbers in a month or two, we may need to update this to handle 3 or 4 numbers properly.

dereknop:
> 1. I wonder if we should put this in dev_scripts (maybe along with adept_repop, jan_one_time, ph1_one_time, and maybe everything that isn't deployment_script? Probably needs a little more discussion). Not necessary though

Sure, np. 

> 2. If I do something crazy like redo.py 0067, it runs 006, which would not be my expected behavior. Maybe we should make sure there are only 3 numbers in that parameter and error otherwise. Also, when we get to bigger script numbers in a month or two, we may need to update this to handle 3 or 4 numbers properly.

Nice catch, I'll fix that.

dereknop:
Okay validation should be much better now

kaitlyn-sharo (approved):

PR #127, opened 2025-04-11
added script to update earlier ta1 session ids in adm history

kaitlyn-sharo:
We had updated ADEPT and apparently ST ADM session ids at some point in time. We changed the session id value in the last `history` entry in `admTargetRuns`, but the TA1 session id also appears in a TA1 Session ID entry and an Alignment Target entry in `history`. Those were not changed. The TA1 Session ID entry was being used to populate the dashboard probe responses page, causing mismatches in data, with the dashboard showing outdated information. 

This script makes sure both of these earlier fields in history are updated to match the correct TA1 session id, found in the last `history` element. 

Run, and check the dashboard and database to see that all three TA1 session ids match.

### DIFF
--- a/scripts/_0_7_0_fix_adm_session_ids.py
+++ b/scripts/_0_7_0_fix_adm_session_ids.py
@@ -1,0 +1,19 @@
+def main(mongo_db):
+    adms = mongo_db['admTargetRuns']
+    eval5_adms = adms.find({'evalNumber': 5})
+    for adm in eval5_adms:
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
Looks good.  The session ID I was looking for appears consistently in the DB and dashboard:
itm-959

brianpippin (approved):
Ran no issues for me.

PR #128, opened 2025-04-11
ITM-957:  Integrated SoarTech alternative alignment into ITM pipeline

nextcen-dgemoets:
SoarTech has been working on an alternative alignment computation "scorecard."
This PR makes this new alignment calculation available such that we can add its results to the dashboard and perform analysis in the future.

To test, run ~`python redo.py 999`~, `python st_newalign_test.py`, observe the use of the API, and check that the alignment obtained using the SoarTech example (and a fixed seed) matches what we get in the Jupyter notebook, 0.9671935933915454.  Let me know if you want to run the notebook yourself to check this value, although you'll need my bugfixes to their example...

~The 999 script won't be merged; it's only for testing this PR.  Although it's~ `python st_newalign_test` is a starting point to the ingest changes that will be needed to support [ITM-961](https://nextcentury.atlassian.net/browse/ITM-961).  Note that the use of the seed is for the purpose of demonstrating that we've converted the Jupyter notebook properly.  We need to determine if we want to use a fixed seed moving forward to get reproducible results, or if we want to leave out the `seed` parameter and let the computed alignment vary somewhat due to random sampling in the calculation.  There is an optional flag to enable random seed generation, off by default.

[ITM-961]: https://nextcentury.atlassian.net/browse/ITM-961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kaitlyn-sharo (approved):
It all looks good to me. This could go in the dev_scripts file, since it's not really something to run in the deployment script (I know you said it wouldn't be merged anyways, so it doesn't matter). I got "does not match" because of a rounding difference.

nextcen-dgemoets:
>  I got "does not match" because of a rounding difference.

Oh interesting.  On Windows, I always get exactly the same (matched) value.

kaitlyn-sharo:
> > I got "does not match" because of a rounding difference.
> 
> Oh interesting. On Windows, I always get exactly the same (matched) value.

Yeah, I figured it was probably a mac problem, or just running it in VSCode rather than the terminal. But it's just the last digit is 5 instead of 4 - not a big deal

brianpippin:
I feel like the 999 script should be renamed and put in the folder Kaitlyn mentioned with some comments about its purpose which could they be used in scripts that need it, so we don't have to delete it.

nextcen-dgemoets:
> I feel like the 999 script should be renamed and put in the folder Kaitlyn mentioned with some comments about its purpose which could they be used in scripts that need it, so we don't have to delete it.

Okay, I moved the 999 script to `git-root/st_newalign_test.py`:
- `python st_newalign_test.py`

I couldn't move it to `dev_scripts` because it uses other modules (like `utils`) but it doesn't have a common parent module from which to import it (without shenanigans).  It seemed like a cleaner solution was to house the test script in the root.

dereknop:
```
SoarTech example:
New VOL Score is 0.9671935933915455.
This does not match the SoarTech Jupyter notebook value.
```

Last digit was a 5 instead of a 4 for me. I assume this is beyond insignificant but figured I'd say something

dereknop:
I had to do another `pip install -r requirements.txt`. I believe that was for sklearn. That may have been my own venv being messed up but this may require an install on prod. Not an issue, just making a note.

nextcen-dgemoets:
> Last digit was a 5 instead of a 4 for me. I assume this is beyond insignificant but figured I'd say something

Yep, same thing happened to Kaitlyn-- we thought it was because she's on a Mac, but I think you're on Windows?  Did you run it from a command line or from VScode (I did cli; she did vscode)?  That was another theory.  Either way, I think it's okay.

dereknop:
> > Last digit was a 5 instead of a 4 for me. I assume this is beyond insignificant but figured I'd say something
> 
> Yep, same thing happened to Kaitlyn-- we thought it was because she's on a Mac, but I think you're on Windows? Did you run it from a command line or from VScode (I did cli; she did vscode)? That was another theory. Either way, I think it's okay.

I ran on Windows. VScode though.

dereknop (approved):

nextcen-dgemoets:
> I had to do another `pip install -r requirements.txt`. I believe that was for sklearn. That may have been my own venv being messed up but this may require an install on prod. Not an issue, just making a note.

Oh yes, I should have mentioned it's probably best to create a new venv from scratch.  And I agree on prod we should either nuke `venv` and re-create it, or it should be fine just to do another pip install as you suggest.

nextcen-dgemoets:
> > > Last digit was a 5 instead of a 4 for me. I assume this is beyond insignificant but figured I'd say something
> > 
> > Yep, same thing happened to Kaitlyn-- we thought it was because she's on a Mac, but I think you're on Windows? Did you run it from a command line or from VScode (I did cli; she did vscode)? That was another theory. Either way, I think it's okay.
> 
> I ran on Windows. VScode though.

That's odd...I just tried running in VScode for the first time, but I got that the values matched.  Oh well...

PR #131, opened 2025-04-22
Clean test runs from 4/21

dereknop:
`python deployment_script.py`

If you grab a copy of the prod database, this script should delete yesterday's two survey test runs. Check localhost:3000/survey-results under legacy and survey version 1.3 to see the change go into effect. 

This script can be tweaked in the future if there are any heavy testing days. All that needs to be changed is the `start_date` and `end_date` variables

### DIFF
--- a/scripts/_0_7_2_wipe_test_data.py
+++ b/scripts/_0_7_2_wipe_test_data.py
@@ -1,3 +1,7 @@
+'''
+Any day where we have a lot of testing before a data collection, this script can be tweaked and run to wipe
+away the noisy data. 
+'''
 def main(mongo_db):
     survey_results = mongo_db['surveyResults']
 
--- a/scripts/_0_7_2_wipe_test_data.py
+++ b/scripts/_0_7_2_wipe_test_data.py
@@ -1,0 +1,14 @@
+def main(mongo_db):
+    survey_results = mongo_db['surveyResults']
+
+    start_date = "Mon Apr 21 2025 00:00:00 GMT-0400 (Eastern Daylight Time)"
+    end_date = "Mon Apr 21 2025 23:59:59 GMT-0400 (Eastern Daylight Time)"
+
+    delete_result = survey_results.delete_many({
+        "$or": [
+            {"results.timeComplete": {"$gte": start_date, "$lte": end_date}},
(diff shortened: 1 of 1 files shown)


brianpippin (approved):
Worked for me.

nextcen-dgemoets:
If this is a script that we might want to re-run, like a utility script, then maybe we should just put it in `dev_scripts` (since it has no dependencies) and not update the DB version afterwards.  We can just update the date in the script and re-run...maybe not even need a ticket/PR/review*2, etc.

Or we can just keep it how it is.  Thoughts, @brianpippin ?

brianpippin:
> If this is a script that we might want to re-run, like a utility script, then maybe we should just put it in `dev_scripts` (since it has no dependencies) and not update the DB version afterwards. We can just update the date in the script and re-run...maybe not even need a ticket/PR/review*2, etc.
> 
> Or we can just keep it how it is. Thoughts, @brianpippin ?

I would say a ticket to create a script that prompts for dates and formats them potentially, that can be run from Jenkins would be a great future item, but for this issue I am fine with it as is.

nextcen-dgemoets:
> I would say a ticket to create a script that prompts for dates and formats them potentially, that can be run from Jenkins would be a great future item, but for this issue I am fine with it as is.

Wrote [ITM-973](https://nextcentury.atlassian.net/browse/ITM-973)

dereknop:
> Before running the script, I see survey 1.3 data on 4/21 for PID 202504000 and 202504001, with a total of 92 records. After running the script, they were gone, and there were 90 records.
> 
> I assume all of the records from 4/22 are legit?

As far as I know, the only test runs were by me and Jennifer on 4/21. So anything after that I am treating as real data.

brianpippin:
> > Before running the script, I see survey 1.3 data on 4/21 for PID 202504000 and 202504001, with a total of 92 records. After running the script, they were gone, and there were 90 records.
> > I assume all of the records from 4/22 are legit?
> 
> As far as I know, the only test runs were by me and Jennifer on 4/21. So anything after that I am treating as real data.

That is my understanding as well.  Anything on 4/22 should be participants until told otherwise.

nextcen-dgemoets (approved):

PR #132, opened 2025-04-23
Results.pid

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?assignee=60ba425da547eb00686ee0ce&selectedIssue=ITM-972)

Kaitlyn and I added the pid to `surveyResults` documents closer to the root level sometime during Phase 1 (`results.pid`). This script will go back through older results and do the same thing for them, making it easier to process them in future ingest tasks. This does not remove the pid from its original location, so anything that points to the previous location on the dashboard will remain unchanged. I also deleted any garbage documents from the collection that were added by cyber. 

`python deployment_script.py`

Then click through some older survey results in Mongo Compass to see that `results.pid` exists

### DIFF
--- a/scripts/_0_7_3_results_pid.py
+++ b/scripts/_0_7_3_results_pid.py
@@ -1,0 +1,50 @@
+'''
+We started placing the pid towards the top level of survey results documents `results.pid`
+This retroactively applies that to older documents 
+Makes it easier for future processing of survey data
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
The number of records in the collection decreased from 580 to 501.
The data looks good.   Moreover, my script for ITM-961 was able to rely on `results.pid` being there.

PR #133, opened 2025-04-23
ITM-961: Add support for new experimental table for SoarTech scorecard alignment

nextcen-dgemoets:
See [ticket](https://nextcentury.atlassian.net/browse/ITM-961) or comments in the script for implementation/algorithm details.  I also updated the new soartech alignment API to have a call that requests/returns all four supported KDMAs, because that's what we want to do here.

To test, I would suggest first changing the script defaults by setting `UPDATE_DATABASE = False` and `VERBOSE_OUTPUT = True`.  Then, run `python deployment_script.py` or `python redo.py 074` if you don't want to update your DB version quite yet.

When you think it's time to update the database, change `UPDATE_DATABASE` and run the deployment script, and check that the DB is update as expected.  Only records with vol scenarios should be updated...but not all such records are updated...

The script flags if there is an unexpected number of comparison records to update in the DB (we expect exactly one).  @dereknop did some sleuthing into these cases.  All but one of the cases where there were no comparison records aren't in the RQ1 table at all.  We'll address these other cases as necessary.


nextcen-dgemoets:
> Only records with vol scenarios should be updated...but not all such records are updated...

I'm not sure why some of these records aren't updated with calibration data, e.g.:
vol-not-updated

dereknop:
> > Only records with vol scenarios should be updated...but not all such records are updated...
> 
> I'm not sure why some of these records aren't updated with calibration data, e.g.: vol-not-updated

Screenshot 2025-04-25 at 9 19 53 AM
I am getting calibration scores attached to all necessary documents. We need the scores on `misaligned` `baseline` and `aligned` documents, and that appears to be the case. You can jump over to ITM-958 on the dashboard side and go to http://localhost:3000/research-results/exploratory-analysis to see the table at the bottom of the page. It is a work in progress but you can see the calibration scores being pulled there.

dereknop (approved):

PR #135, opened 2025-05-16
3.9 compatibility

dereknop:
Wipe away db, replace with a backup.

switch your venv to using python 3.9

check the calibration scores script ran correctly

### DIFF
--- a/utils/golden_arm.py
+++ b/utils/golden_arm.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pandas as pd
 from os import PathLike
-from typing import Optional, Iterable
+from typing import Optional, Iterable, List, Dict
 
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets:
@brianpippin Do we need to do a CACI TRADE SECRET-like thing since this is SoarTech code we've modified?

dereknop:
> @brianpippin Do we need to do a CACI TRADE SECRET-like thing since this is SoarTech code we've modified?

if it matters at all, I am ok with getting no credit for minorly changing two function signatures

brianpippin:
I'm not worried about it here because all this code was developed on ITM and goes to the customer anyways.

nextcen-dgemoets (approved):
LGTM.

PR #136, opened 2025-05-30
Itm 1007

dereknop:
Converts the YAML files into JSON configs that can be used for the text-based scenarios.

`python deployment_script.py`

Then test the [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/295)


nextcen-dgemoets:
Looks good.  ADEPT submitted their updates to the Personal Safety dataset, so I have three new PS YAMLs for you.  I'll send them to you in Teams.

nextcen-dgemoets:
What is `template Probe N` in the `questions` under `Probe N` of each object?  It looks empty.
Related, What is `probe` under `Probe N` -> `questions` -> `probe Probe N` (and `template Probe N`?  Also looks empty.

nextcen-dgemoets:
I tried repeating the survey with the same email address / PID.  It created 4 new entries in the DB, including a second MF2 because I believe it was randomly selected both times.  Is this the expected behavior?

nextcen-dgemoets:
Oh also-- are we doing multi-kdma scenarios?

PR #138, opened 2025-06-10
Add june 2025 scenarios to dashboard

dereknop:
This is the same type of script we have had to do for all of the evaluations (see script _0_6_4_, _0_3_4_).

`python deployment_script.py`

Check the `scenarios` collection using MongoDB Compass to ensure the scenarios have been added

### DIFF
--- a/phase2/june2025/june2025-adept-eval-AF-MF1.yaml
+++ b/phase2/june2025/june2025-adept-eval-AF-MF1.yaml
@@ -1,0 +1,367 @@
+id: June2025-AF-MF1-eval
+name: Affiliation Focus Set A Merit Focus Set A 1
+state:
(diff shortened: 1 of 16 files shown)


dereknop:
> What happens if there are scenarios added here but there's no ADM data for these scenarios? I'm just wondering if we should add all the scenarios as part of this ingest ticket instead of adding the rest of them in another ticket.

Nothing bad happens, the only ones I didn't add are the training scenarios and the non-subset files. I can add in the non subset ones.

nextcen-dgemoets (approved):
LGTM.  They look like the right scenarios and look good in the DB.

brianpippin (approved):
Ingested and worked for me.

PR #139, opened 2025-06-11
p2_j_2025

dereknop:
Use most recent backup I generated this morning. http://localhost:3000/results, no performer/adm will show up when you select a scenario.
`python deployment_script.py`

ADM's should now show up. Note that some will still be missing because they haven't run (e.g `AF1-MF1-eval`)


nextcen-dgemoets:
This PR was closed because there is a Jenkins job that can redo a script.  In this case, we'll run it on 051 when the ADMs are done running.

PR #140, opened 2025-06-13
Delegation Materials and Text Scenarios (updated)

dereknop:
First, grab the most up-to-date backup from S3.

`python deployment_script.py`

This will regenerate the text-based scenarios with the changes to separate the common text vs. the probe-specific text. It will also generate delegation materials for each ADM run (excluding the `prodtest` runs). Those materials all get uploaded to the `admMedics` collection. There is one more step after that to upload the new survey config.

`cd delegation_survey`
 `python update_survey_config.py`
`enter`
`6`
`y`
`y`

Now you can use the dashboard pr to look at http://localhost:3000/review-delegation


nextcen-dgemoets:
These are (as far as I know) the current ADMs that don't need delegation materials and should be ultimately be deleted:
- `adm_name` = "prodtest" or "prod_multitest"
- `adm_name` contains "Random"
- `adm_name` contains "6d0829ad-4e3c-4a03-8f3d-472cc549888f"

text_based_scenarios/convert_yaml_to_json_config_ph2.py
```diff
@@ -118,7 +120,9 @@ def main(mongo_db):
 
     for filename in os.listdir(phase2_folder):
         file_path = os.path.join(phase2_folder, filename)
-
+        # don't gen multi kdma 
+        if "AF-MF" in filename:
```
nextcen-dgemoets, line 124: Why aren't we getting multi-kdma yet?

nextcen-dgemoets (commented):
Preliminary review / comments.

> this will need to be replaced with if scene[id] in list of selected probes (i dont have yet)

What are you missing here?  Is this something that's available yet?

Note the ADM names I listed in a previous comment that should be excluded.

dereknop:
> Preliminary review / comments.
> 
> > this will need to be replaced with if scene[id] in list of selected probes (i dont have yet)
> 
> What are you missing here? Is this something that's available yet?
> 
> Note the ADM names I listed in a previous comment that should be excluded.

This is an outdated comment I left for myself I should delete. This is from before I had subset files and thought I would just be given a list of the probe ids we were using for the actual text-based scenarios

text_based_scenarios/convert_yaml_to_json_config_ph2.py
dereknop: I am not aware of us using the Multi-KDMA files for text-based scenarios yet. If we are and they need to be incorporated, I can just remove this and this script will convert them like the rest of the yaml files. I have the YAML. files here so they are handy for the adm conversion script, but they are currently not used in the text-based conversion script.

brianpippin:
The ingest ran with no issues for me, I want to test some stuff in the dashboard before I approve this.  Future ticket, I would love to figure out how to add the update_survey_config stuff to a script at some point or bash or something so that we don't need to sign onto the machine and run those commands.

joshuale-caci (approved):
Script ran fine after doing the redo script and everything looks right on my dashboard

PR #141, opened 2025-06-16
ITM 1030

dereknop:
http://localhost:3000/results

You'll see some test results polluting our data. After running:

`python deployment_script.py` 

all of these test runs should be removed. 

Known test runs were any adm names that included `test`, `random`, or `6d0829ad-4e3c-4a03-8f3d-472cc549888f`


nextcen-dgemoets (approved):
> Approving. Did some research and the "undefined" that Darren noted I believe is not tied to adm_names but some other underlying issue we need a ticket for.

By that same token, I will approve as well.

PR #142, opened 2025-06-17
ITM-1013

dereknop:
http://localhost:3000/participant-progress-table

Click on phase 2. You'll see the extra data populated here from 6/2 and 6/3. Any pid before `202506113` is testing and needs to be deleted. I am deleting the text documents themselves as well as the entries in the participant log so that they will be removed from this table (and their data won't leak into other tables across the dashboard). I am also deleting participant log entries that had 0 text responses (there were only two cases of this). This happens when the page was loaded in and none of the scenarios were actually completed.

`python deployment_script.py`

Check that all of the rows in the participant progress table are now from actual participants.

### DIFF
--- a/scripts/_0_8_1_clean_text_results_and_plog.py
+++ b/scripts/_0_8_1_clean_text_results_and_plog.py
@@ -1,0 +1,62 @@
+def main(mongo_db):
+    p_log = mongo_db['participantLog']
+    text_results = mongo_db['userScenarioResults']
+    
+    min_pid = 202506100
+    max_pid = 202506112
+    
+    text_query = {
+        "participantID": {
+            "$gte": str(min_pid),
+            "$lte": str(max_pid)
+        }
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  The program output said which IDs/PIDs were being removed (and why).  Comparing the downloaded excel file of the progress log before and after the ticket showed the expected rows removed from the progress log.

I did notice (unrelated to this ticket) that the download for the phase 2 progress table includes columns for phase 1 scenarios, e.g., QOL1, MJ2, etc.  Likewise, the phase 2 progress table includes columns for the phase 2 scenarios, plus QOL4 and VOL4, that I don't think exist.

brianpippin (approved):
Worked for me.

joshuale-caci (approved):
Deleted all the right participant ids

PR #145, opened 2025-06-24
Itm 973

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-973)

This script takes a date as a parameter and then deletes any data from that day from `userScenarioResults`, `surveyResults`, and `participantLog`. This should help delete test data moving forward. I will point a Jenkins job to it, similar to the one for the redo script, once the behavior is verified as being correct. 

To test run this command from the root directory:
`python clean_data_from_date.py xx/xx/xxxx`

I think the easiest way to view the deletion of documents would be using the participant progress table. The participant IDs being removed will be logged in the terminal.

### DIFF
--- a/clean_data_from_date.py
+++ b/clean_data_from_date.py
@@ -1,24 +1,136 @@
+'''
+https://nextcentury.atlassian.net/browse/ITM-973
+This script should be used when prod needs to be cleaned of testing data from a certain date. 
+Usage ex. `python clean_data_from_date.py 06/24/2025`
+It will clean out the surveyResults, userScenarioResults, and participantLog collections from that date.
+'''
 import sys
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
Worked for me.  It didn't clear out one of the rogue PIDs in the progress log on prod (`ALS_test1210b`), but that one never got as far as finishing (or starting) the text scenarios (Text Start Date is empty).  There's probably nothing to be done there, unless anybody has any ideas.

jaudick (approved):
The objects from the dates I entered were deleted

PR #146, opened 2025-06-25
ITM-1038: Create random subsets from ADM results of June2025 Collaboration

nextcen-dgemoets:
Uses `probe_midpoint` field from the ADEPT csvs and the thresholds established by them to pick probes from the various “bins” and create 100 randomly-generated scenarios per kdma.  Look up ADM results for every probe in the database (at each alignment target, both aligned and baseline) and calculate new alignment and kdmas for the synthetic scenarios.  Create new synthetic entries in the `admTargetRuns` collection as if the ADMs had actually run through these subset scenarios.

I think it's easier for testing purposes to run it from the command line, especially since I've added some verbose output that helps show what it's doing:
`python scripts/_0_8_2_generate_random_adm_probesets.py -h`
```
Generates ADM results for synthetically created random assessment subsets.

options:
  -h, --help            show this help message and exit
  -v, --verbose         Verbose logging
  -n, --no_output       Do not write to the MongoDB
  -c count, --count count
                        How many scenarios to generate per KDMA (default 100)
  -i ignore [ignore ...], --ignore ignore [ignore ...]
                        Acronyms of attributes to ignore (AF, MF, PS, SS, AF-MF)
```

So try something like `python scripts/_0_8_2_generate_random_adm_probesets.py -v -n -c 3` to get an idea what it will be doing.  Note that this still hits the TA1 server, so consider running against your local TA1 server with unredacted scenarios (and updating `.env` appropriately).  When you're testing writing to the db, you might temporarily want to change the collection name in `create_synthetic_adm_runs` to `admTargetRunsNew` so that you're not updating your `admTargetRuns` table, and you can drop/re-create the test collection for subsequent tests.

Things to test/verify:
- The bins are being correctly constructed based on the TA1 ranges (see the Word document I put in dev chat)
- The synthetic scenarios are randomly choosing a probe from each subset, including for multi-kdma
- Alignment and kdmas are properly being requested by TA1 and stored
- The synthetic ADMs look good/correct in the `admTargetRuns` database

But ultimately, we'll need to run:
`python deployment_script.py`


dereknop (approved):
This worked as you described for me, and nothing on the dashboard broke. 

Screenshot 2025-06-25 at 12 32 10 PM

jaudick (approved):
I got it all to work

nextcen-dgemoets:
> Everything seemed to work but when I ran against my local adept server I wasn't able to get scores and ran into this error

I'm guessing you didn't have the unredacted scenarios in your TA1 server.  It can't generate a score from redacted scenarios, which is currently what's in the ADEPT repository.  That said, I could have more gracefully recovered from that...

PR #150, opened 2025-07-24
Itm 1066

dereknop:
Converts the new ADM runs into delegation materials and performs survey setup. 

You need a Mongo backup from after the ADM runs yesterday. Let me know if you need one. I edited some existing scripts so that nobody needs to `cd` into a different directory and run separate commands anymore. You should be able to just run:
`python deployment_script.py`. This way, our Jenkins job is all we will need to deploy properly. 

After the script runs, checkout the [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/317).


NeilDaniel07 (approved):
Everything looks good to me. Running the deployment script resulted in my database version being updated to 0.8.5. Success statements were printed to the terminal.

PR #151, opened 2025-07-28
AF3 Fix

dereknop:
`python deployment_script.py`

This will delete the existing ADM runs for AF3, clean the amd_name field of the new ones, convert them into medics to be used in the survey (deleting old medics), and finally generate an updated survey config. It also replaces the AF3 text scenario file with the newly updated one. 

You can quickly do a spot check on the new AF3 file here http://localhost:3000/review-text-based to make sure it has 6 probes.

After you run this, check out the [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/319) for the new survey.


dereknop:
> All of the files LGTM. After redoing 086, my `surveyConfig7x.json` has diffs. Should this be checked in?

Not necessary. The diffs are coming from Medic names being assigned to different scenario/adm/target combinations, not from actual content changes (Besides the change to AF3 runs, which I checked in yesterday).

nextcen-dgemoets (approved):

dereknop:
> Script ran fine and AF3 showed with 6 probes. It did however tell me it had no incorrect AF3 runs but checking my database showed I had none to match the case so it might be an old database.

In that case you should probably grab a backup. I'll put one on teams

PR #152, opened 2025-07-30
Add eval numbers to participant log

kaitlyn-sharo:
Run the deployment script, then check out your participant log. This should add eval numbers to all the participants in the participant log.

### DIFF
--- a/scripts/_0_8_7_add_evals_to_plog.py
+++ b/scripts/_0_8_7_add_evals_to_plog.py
@@ -1,0 +1,33 @@
+def main(mongo_db):
+    plog_collection = mongo_db["participantLog"]
+    all_docs = plog_collection.find({})
+    dre = [202409100, 202409199] # 4
+    alsoDre = [20249201, 20249299] # 4
+    ph1 = [202411300, 202411399] # 5
+    ph1Online = [202411500, 202411599] # 5
+    janEval = [202501700, 202501799] # 6
+    eval8 = [202506100, 202506299] # 8
+    eval9 = [202507100, 202507299] # 9
+    for doc in all_docs:
+        if 'hashedEmail' in doc:
+            pid = doc['ParticipantID']
+
+            try:
+                pid = int(pid)
+            except:
+                print(f'Could not convert pid {pid} to int')
+                continue
+            evalNum = None
+            if (pid > dre[0] and pid < dre[1]) or (pid > alsoDre[0] and pid < alsoDre[1]):
+                evalNum = 4
+            elif (pid > ph1[0] and pid < ph1[1]) or (pid > ph1Online[0] and pid < ph1Online[1]):
+                evalNum = 5
+            elif (pid > janEval[0] and pid < janEval[1]):
+                evalNum = 6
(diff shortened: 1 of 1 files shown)


NeilDaniel07 (commented):
This seemed to only partially work for me. When I ran `python deployment_script.py`, I saw the message that my database version has successfully been updated. However, I saw that it did not work for two PIDs, because they were not integers:
`Could not convert pid 202409113A to int`
`Could not convert pid 202409113B to int`

While I see that you handled this in a try-catch block, I'm wondering what your thoughts are on parsing the numerical portion of the PID using Regex and then converting the result to an `int` instead of just trying a raw conversion (I would assume that the A and B in these PIDs don't affect the `evalNumber`). This might be more resilient against these edge cases and future ones. 

Other than that, when I ran the query `{ evalNum: { $exists: true } }` in Mongo compass, I saw that only 269 or 291 documents in the collection were selected for my database version. I'm not sure if the other 20 documents (excluding the two mentioned above) are not receiving evalNum information because of some other edge case.

scripts/_0_8_7_add_evals_to_plog.py
```diff
@@ -0,0 +1,33 @@
+def main(mongo_db):
+    plog_collection = mongo_db["participantLog"]
+    all_docs = plog_collection.find({})
+    dre = [202409100, 202409199] # 4
+    alsoDre = [20249201, 20249299] # 4
+    ph1 = [202411300, 202411399] # 5
+    ph1Online = [202411500, 202411599] # 5
+    janEval = [202501700, 202501799] # 6
+    eval8 = [202506100, 202506299] # 8
+    eval9 = [202507100, 202507299] # 9
```
nextcen-dgemoets, line 10: I know it's just a name thing, but can we call these `june2025` and `july2025` instead of `eval8` and `eval9`?  They deserve descriptive names like the variables above them.     :)

kaitlyn-sharo: Sure, but only because I have to fix the other problem ;)

kaitlyn-sharo:
- Updated to change < and > to <= and >=, respectively. 
- Handled the A/B edge case, which I had planned to do, but got called away before I remembered
- Updated the names of eval8 and eval9 to juneEval and julyEval

There are some participants without hashedEmails - these participants will not get evalNumbers. The evalNumbers are only being added to help with duplicate email handling in the dashboard. However, if there is an evalNumber that should be added to these (I noticed 7 is missing?), I am happy to add that, just for completeness.

scripts/_0_8_7_add_evals_to_plog.py
```diff
@@ -0,0 +1,36 @@
...
+            pid = doc['ParticipantID']
+
+            try:
+                pid = int(pid)
+            except:
+                if pid == '202409113A' or pid == '202409113B':
+                    evalNum = 4
+                else:
+                    print(f'Could not convert pid {pid} to int')
+                continue
```
dereknop, line 22: I think this `continue` is stopping the two PIDs that Neil mentioned from getting updated.

dereknop (approved):

PR #153, opened 2025-07-31
Itm 1035 rq8 updates

kaitlyn-sharo:
To run: `python3 june2025_probe_matcher.py -i ph2_sim_files`

To see the values that populate, check RQ8 in the exploratory analysis table on the dashboard (you can do that on development or on the branch I'm going to be pushing relatively soon. The data is the same either way).

Updates:
- Added hemorrhage control - a binary value that says whether a patient completed hemorrhage control or not. To see which injuries on each patient requires hemorrhage control, see REQ_HEMORRHAGE_PROCEDURES at the top of the file. 
- Added time to hemorrhage control - the time from the start of the simulation (SESSION_START) until the final hemorrhage procedure is completed. If hemorrhage control is incomplete, this is `None`. Formatted in mm:ss based on the variable definitions sheet for RQ8
- Added triage performance. 
    - For the injuries on each patient, there is one correct matching treatment (i.e. chest puncture -> chest seal; shoulder puncture -> hemostatic gauze (aka woundpack); broken bone -> splint; etc.). The csv handles this in the INJURY_TREATED line. If the correct treatment is given, we add one to `correct_tools_applied`. 
    - When any tool is applied (`TOOL_APPLIED`), we add one to `total_tools_applied`. 
    - If the tool applied is an approved supplemental tool (as determined by Nick from OSU), we add that to a temporary count of supplemental tools _per patient_
    - At the end, if any required treatments have not been applied, we add that to a `misses` variable
    - For each patient, if the participant applied all required treatments, we add the supplemental tools to the `correct_tools_applied` variable. This was Jacob Hyde's idea
    - The triage performance is calculated as `correct_tools_applied / (misses + total_tools_applied)`

### DIFF
--- a/june2025_probe_matcher.py
+++ b/june2025_probe_matcher.py
@@ -573,7 +573,7 @@ class ProbeMatcher:
                 if patient not in to_complete:
                     correct_tools_applied += supplemental_points[patient]
 
-            return correct_tools_applied / max(1, total_tools_applied + misses)
+            return (correct_tools_applied / max(1, total_tools_applied + misses)) * 100
         
         results[f'{env} Triage Performance'] = get_triage_performance()
 
--- a/june2025_probe_matcher.py
+++ b/june2025_probe_matcher.py
@@ -515,10 +515,6 @@ class ProbeMatcher:
             # if there are no patients left in the dictionary, we have succeeded with hemorrhage control!
             res = 1 if len(list(to_complete.keys())) == 0 else 0
             time_to_hcontrol = (end_time-start_time) if res == 1 else None
-            if time_to_hcontrol:
-                minutes = int(time_to_hcontrol // 60)
-                seconds = int(time_to_hcontrol % 60)
-                time_to_hcontrol = str(minutes) + ':' + str(seconds)
             return {'completed': res, 'time': time_to_hcontrol}
         
         hem_control = get_hemorrhage_control()
--- a/june2025_probe_matcher.py
+++ b/june2025_probe_matcher.py
@@ -1,4 +1,4 @@
-import yaml, argparse, json, os, csv, requests
+import yaml, argparse, json, os, csv, requests, copy
 from logger import LogLevel, Logger
 from pymongo import MongoClient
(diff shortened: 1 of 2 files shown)


june2025_probe_matcher.py
```diff
@@ -416,6 +488,99 @@ def get_ps_ss():
...
+                                if len(to_complete[patient]) == 0:
+                                    del to_complete[patient]
+
+            # if there are no patients left in the dictionary, we have succeeded with hemorrhage control!
+            res = 1 if len(list(to_complete.keys())) == 0 else 0
+            time_to_hcontrol = (end_time-start_time) if res == 1 else None
+            if time_to_hcontrol:
+                minutes = int(time_to_hcontrol // 60)
+                seconds = int(time_to_hcontrol % 60)
+                time_to_hcontrol = str(minutes) + ':' + str(seconds)
```
nextcen-dgemoets, line 521: I'm confused about the time format(s), particularly in what is desired here.  In the ticket, Jennifer seems to want the triage time in milliseconds (ms)...unless that's supposed to mean mm:ss (as in minutes:seconds)?  Then for hemorrhage time, she wants it in minutes (m), though maybe that's supposed to be mm:ss as well?

Either way, the time formats are different (between hemorrhage control and triage time), and maybe they should be the same.  And maybe we clarify the time format that Jennifer wants.  She didn't object to the triage time being e.g. 486.974.  But I'm not sure if she wants all times to be in mm:ss or what.  And even if it's mm:ss, does she want/require e.g. 4:09 instead of 4:9 (which is how it is currently), or does she care?

nextcen-dgemoets (commented):
Looks good in ingest.  My only real question is about desired time-format (see below).

june2025_probe_matcher.py
kaitlyn-sharo: Discussed with Jennifer. Time is now in seconds.

joshuale-caci (approved):
I got my tables updated and they show the data now!

PR #157, opened 2025-08-06
Repopulation and Comparison scores

dereknop:
[Repopulation ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-1085)
[Redo script tweak ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce&selectedIssue=ITM-1083)

Set up your `.env` file to point to your local ADEPT server. For the repopulation script to work properly, you will need to have all of the July and June YAML files accounted for. 

`python deployment_script.py`

This will take a few minutes. After running, the session IDs for human results, as well as the ADM runs, should be stored on your local ADEPT server. Now we can test the changes to the redo script while also generating some comparison scores. 

I realized that there are certain instances where we will need to be able to pass an argument when running `redo.py`. For example, script `083` generated the June comparison scores for delegators against the observed ADMS. That script has an `EVAL_NUMBER` parameter (with default value of 8), but will work the same for eval number 9. In this case, we can use our Jenkins job for the redo script, but it needs to be capable of passing a new `EVAL_NUMBER`.

Changes: 
- Minor tweak to `083` in a few spots that had eval 8 specific logic hard-coded
- redo.py can take additional arguments from the command line
- redo.py determines what type the argument should be treated as (int, float, bool, string)

A quick way to test this would be to run `python redo.py 083 9`. The redo script should identify that we need to run script 083, and then pass 9 (as an int) as the `EVAL_NUMBER`. You should then be able to check `humanToADMComparison` with  `{ evalNumber: 9 }` and see documents that were not previously there. Let me know if you need the latest db backup to test this.

Note: Jenkins job update accompanies this, but is not part of reviewing this pr.

### DIFF
--- a/ph2_repop.py
+++ b/ph2_repop.py
@@ -14,14 +14,6 @@ import utils.db_utils as db_utils
 MONGO_URL = config("MONGO_URL")
 ADEPT_URL = config("ADEPT_URL")
 
-import requests
-from decouple import config
-from collections import defaultdict
-import utils.db_utils as db_utils
-
-MONGO_URL = config("MONGO_URL")
-ADEPT_URL = config("ADEPT_URL")
-
 
 def main(mongo_db):
    text_scenarios = mongo_db["userScenarioResults"].find({"evalNumber": {"$gte": 8}})
--- a/ph2_repop.py
+++ b/ph2_repop.py
@@ -14,68 +14,77 @@ import utils.db_utils as db_utils
 MONGO_URL = config("MONGO_URL")
 ADEPT_URL = config("ADEPT_URL")
 
+import requests
+from decouple import config
+from collections import defaultdict
+import utils.db_utils as db_utils
+
+MONGO_URL = config("MONGO_URL")
+ADEPT_URL = config("ADEPT_URL")
+
 
 def main(mongo_db):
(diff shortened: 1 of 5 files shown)


nextcen-dgemoets:
I got a bunch of "bad scenario/probe" type errors on my local TA1 server at first.  I can't tell which ones they are.  Is it trying the OW scenarios?

dereknop:
> I got a bunch of "bad scenario/probe" type errors on my local TA1 server at first. I can't tell which ones they are. Is it trying the OW scenarios?
> 
> Other than that, everything looked good.

Yes, I would expect those at the beginning of the run since the first June text-based scenarios will be processed first. This gets handled with my change in `db_utils.py`. It will retry with the appropriate probe it pulls out from the full set instead of the subset.

nextcen-dgemoets (approved):
LGTM. The deployment script ran 089, and it was hitting my local TA1 server.  I mentioned the errors it gave, but you already addressed that in comments.

I ran the redo script with `083 9` as instructed and ended up with 384 eval 9 documents in `humanToADMComparison`  I didn't see any anomalies in these entries.

I also saw the improved ADM names in `/adm-probe-responses`.

PR #158, opened 2025-08-11
String to int eval num

dereknop:
`python deployment_script.py`

http://localhost:3000/results

Should see full set runs now. Should only be one 'Phase 2 July 2025 Collaboration"


nextcen-dgemoets (approved):
It works, and now my random subset script works, too.

NeilDaniel07 (approved):
Looks good to me. I now only see one `Phase 2 July 2025 Collaboration` tab in the dashboard. The table and ADM run information is now available for full sets.

PR #159, opened 2025-08-11
ITM-1091: Create random subsets from ADM results of July2025 Collaboration

nextcen-dgemoets:
Updated `_0_8_2_generate_random_adm_probesets.py` for July.

This PR requires [Derek's recent PR](https://github.com/NextCenturyCorporation/itm-ingest/pull/158) to be run/deployed on your local database.

For testing, you might want to set `NUM_SUBSETS` to something smaller, like 3.  Or generate the full 100 subsets, as you see fit.

Then run `python redo.py 082`.  Subsequent calls to 082 will replace the synthetic scenarios that were previously in the DB.

Check mongo to see that `NUM_SUBSETS * 96` synthetic scenarios have been added to eval 9.


NeilDaniel07 (approved):
Worked exactly as described in the testing instructions. I set `NUM_SUBSETS` to 3 and saw that the script ran successfully for me. When I checked Mongo compass, I saw that 96*3 = 288 new documents were present for the query `{ evalNumber: 9 }`.

dereknop (approved):
Screenshot 2025-08-12 at 9 50 46 AM
I ran 3 subsets and got 288 synthetic subsets, which checks out. I used my local ADEPT server because hitting prod was taking forever, even with only 3 subsets. LGTM. Furthermore, http://localhost:3000/research-results/rq2 is looking good. I expect that when this is run on production, the table will auto-populate without an update being needed.

PR #160, opened 2025-08-18
Uploaded sim data up through 8/17

kaitlyn-sharo:
This has all the sim data for Phase 2 through August 21. It also includes a minor code change to the june2025 probe matcher so it can run june and july open world scenarios at the same time. 

Everything should run smoothly (no errors/warnings) and you should see the data updated in RQ8.

Make sure you have an updated database to test!

`python june2025_probe_matcher.py -i ph2_sim_files`

### DIFF
--- a/june2025_probe_matcher.py
+++ b/june2025_probe_matcher.py
@@ -797,7 +797,7 @@ if __name__ == '__main__':
             EVAL_NAME = 'July 2025 Collaboration'
             EVAL_PREFIX = 'june2025'
             EVAL_NUM = 9
-        elif "202507" in str(pid):
+        elif "202509" in str(pid):
             EVAL_NAME = 'September 2025 Collaboration'
             EVAL_PREFIX = 'june2025'
             EVAL_NUM = 10


dereknop (approved):

june2025_probe_matcher.py
```diff
@@ -745,12 +745,8 @@ def first_engaged(characters: list):
     if args.is_verbose:
         VERBOSE = True
     EVAL_NUM = DEFAULT_EVAL_NUM if not args.eval_num else args.eval_num
-    if EVAL_NUM == 8:
-        EVAL_NAME = 'June 2025 Collaboration'
+    if EVAL_NUM == 8 or EVAL_NUM == 9:
```
nextcen-dgemoets, line 748: Why does an `EVAL_NUM` of `9` map to `june2025`?  This appears deliberate, so there must be a reason.  I just can't think off-hand what it is.  That said, I haven't caught up on Dashboard chat.  Maybe we're just considering June == July for sim/RQ8 purposes...except for mystery PID `202507`, who gets to be July.

kaitlyn-sharo: Yes, june == july. We discussed it and since june, july and september will all be the same open world scenarios, we only need the june yamls, so we'll just stick with june - no need to separate anything. Ideally, we'd change everything to "phase2", in my opinion, but I didn't want to mess with anything that could've already been used somewhere else

PR #163, opened 2025-08-27
removed incorrect open world entries from ph2 database

kaitlyn-sharo:
In the participant progress table, phase 2 participants were showing up with 3 sims: desert, urban, and eval_open_world. eval_open_world was for phase 1 only.

This ticket removes all the eval_open_world data that's already been stored in the database (run the deployment script) and ensures it does not get re-added when the probe matcher is run.

It also moves the action analysis data and text kdmas into the desert and urban documents, rather than having them in the eval_open_world document.

Run the deployment script and the probe matcher, then check out the corresponding dashboard PR to see that only the phase 2 open world sims are listed in the progress table.

`python deployment_script.py`
`python june2025_probe_matcher.py -i ph2_sim_files -e 9`

### DIFF
--- a/june2025_probe_matcher.py
+++ b/june2025_probe_matcher.py
@@ -615,11 +615,15 @@ class ProbeMatcher:
 
         self.logger.log(LogLevel.INFO, f"{'' if SEND_TO_MONGO else 'NOT '}Saving to database.")
         if SEND_TO_MONGO:
-            mongo_id = self.participantId + f'_{EVAL_PREFIX}_open_world'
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets:
> Sorry for requesting rereview, Darren. Please see the comment on the dashboard to see why I needed to make this additional change

Wait a minute-- now I'm not seeing any of the OW analysis fields in RQ8-- was that intended?  Maybe I need to grab a clean DB.  This feels like user error.

jaudick (approved):

PR #167, opened 2025-09-17
ITM-1126: Update Phase 1 2d dev script for Phase 2 4d

nextcen-dgemoets:
Take the 2d dev script from Phase 1 (MJ / IO) and update it for Phase 2 4d (AF, MF, PS, SS).

Run `python dev_scripts/get_text_kdmas.py` and examine the resulting `text_kdmas.csv` file to see that it looks properly formatted with the Eval 9 (July) participants' KDMAs.  Spot check against your Mongo DB.

Then continue on to the [server PR](https://github.com/NextCenturyCorporation/itm-evaluation-server/pull/151).


kaitlyn-sharo (approved):
Ran, spot-checked, looked good to me! I do not see the extra empty lines that Neil is seeing. Just curious about whether we want to keep phase 1 and 2 separate, either within the same file or different files, like we discussed on teams. but in terms of functionality, no problems here!

nextcen-dgemoets:
> My only trivial comment is regarding the spacing between rows of the CSV file. Currently, only every second row is populated, leaving all even numbered rows blank. I'm not sure if this is intentional nor do I think the effect will be significant for operating on the file in Excel/Google Sheets/etc, however I believe that the file would match expected standards more closely if every row was populated. I just wanted to bring this to your attention, but can approve these changes in their current state.

This is a carriage return / line feed  *nix vs. Windows thing, which is why Kaitlyn doesn't see it.  If you bring it up in an (intelligent) editor, it looks fine.  If you have a cross-platform fix for it, I'd be happy to update the script.  But what I do now is skip empty (and other invalid) lines on the server side.

nextcen-dgemoets:
> Just curious about whether we want to keep phase 1 and 2 separate, either within the same file or different files, like we discussed on teams. but in terms of functionality, no problems here!

Based on internal and stand-up conversation, I've made this script separate from the phase 1 script, and renamed the latter accordingly.

brianpippin:
"Wrote 52 rows to text_kdmas.csv", if that is correct number then I think we are good.

joshuale-caci (approved):
My database showed 56 rows. Doubled checked with the pid specific values in the database and they line up correctly with the values in the csv.

PR #168, opened 2025-09-18
Update participant mistakenly marked as 'Test'

dereknop:
[Ticket](https://nextcentury.atlassian.net/jira/software/projects/ITM/boards/116?jql=assignee%20%3D%2060ba425da547eb00686ee0ce)

`python deployment_script.py`

This updates participant `202507147` to be included in our data sets. This participant was somehow marked as type `Test` instead of `emailParticipant`. I set up this script so it can be run again with the Jenkins redo job if this happens in the future (taking the PID that needs to be updated as an argument)

### DIFF
--- a/scripts/_0_9_7_fix_test_log.py
+++ b/scripts/_0_9_7_fix_test_log.py
@@ -1,0 +1,7 @@
+def main(mongo_db, pid=202507147):
+    participant_log = mongo_db['participantLog']
+
+    participant_log.find_one_and_update(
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  I ran the deployment script and pid `202507147` was fixed to `emailParticipant`.  Then  I ran `redo.py 097 202509302` and it moved pid `202509302`, which was previously one of my test pids, to `emailParticipant`.

PR #169, opened 2025-09-18
Itm 1127 september files

kaitlyn-sharo:
Run `python june2025_probe_matcher.py -i ph2_sim_files`

Check the dashboard participant progress table to see the sim columns populate for september


dereknop:
> If you have an updated database, there should be no errors. There were a handful due to 2025093## being labelled as eval 9 instead of 10 - that is fixed now. I have removed all the tutorial levels and alpha squad. Now it should run perfectly clean

2025093## were not part of the September collection (Eval 10). Their text-based scenarios are all marked as eval 9 since we used the same materials as the July Collaboartion. 

20250910xx -> Septmeber collab
2025093xx -> Continuation of July materials from last weekend

dereknop (approved):
Ran with no errors

PR #172, opened 2025-09-24
ITM-1134:  Phase 2 4K experiment

nextcen-dgemoets:
This is based on script 082 which is the Phase 1 equivalent, for comparison purposes.

You need the database from at least Monday night when the ADMs were complete.
You **also need to run** the 099 script from ITM-1137-- see [the open PR](https://github.com/NextCenturyCorporation/itm-ingest/pull/171).
You can run the deployment script, or run it directly:
- `python scripts/_1_0_0_p2_4k_experiment.py`

On my computer at least, it's taking a long time to get all of these comparison alignments, so you might try calling the script directly with `-c 3` so that there are 3 instead of 30 subsets.

Check the `admTargetRuns` collection in your database where `evalNum` is 11 and `synthetic` is true.

In `evaluation`, you should see human kdmas (which are the kdmas the ADM was trying to align to).
In `results`, you should see the alignment score and kdmas for that particular synthetic ADM run.
In the root of the collection, you should also see the `probe_ids` for the randomly-generated subset.

### DIFF
--- a/scripts/_1_0_0_p2_4k_experiment.py
+++ b/scripts/_1_0_0_p2_4k_experiment.py
@@ -58,14 +58,15 @@ all_adm_data: dict = {}
 class Adm_data:
     adm_name: str
     text_session_id: str
+    human_scenario_id: str
     human_kdmas: list
     pid: str
     scenario_id: str
     probe_responses: dict
 
 
-# Get TA1 session ID from text survey with the target kdma values of the specified ADM run
-def get_text_session_id(mongo_db, adm_entry):
+# Get various data from text survey with the target kdma values of the specified ADM run
+def get_text_survey_info(mongo_db, adm_entry):
     # First, get the KDMA values from the ADM's alignment target-- which originally comes from the human taking the text survey.
     target_text_kdmas: dict = {}
     for history_entry in adm_entry["history"]:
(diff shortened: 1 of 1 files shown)


dereknop:
Is it expected that a different number of probes can be added ot each bin? I don't know how the bin logic works.
Screenshot 2025-09-26 at 10 56 26 AM

dereknop:
I see the results where `evalNumber: 11` and `synthetic: true`

Screenshot 2025-09-26 at 11 02 45 AM

I can see the pid the run is aiming at and the resulting alignment score. This all looks good to me. 

If I query `{ evalNumber: 11, synthetic: { $exists: false } }` I do get 488 results and I am curious what these runs are exactly. 

Also, I wanted to note that with this pointed at my local ADEPT server, I think it ran in a reasonable amount of time. Is this potentially a local db issue for me that I got 3480 synthetic runs for SS but not the others? Or is that not an issue to be concerned about?
Screenshot 2025-09-26 at 11 04 32 AM

nextcen-dgemoets:
> Is it expected that a different number of probes can be added ot each bin? I don't know how the bin logic works. 

Yes, because there are different numbers of probes in each bin.  For example in AF, the first bin contains probes with midpoint less than 0.20, and there are 10 such probes in the full list of AF probes.  The next bin contains probes with midpoint from .20 to .34999, but there are only 6 such probes in the full list of AF probes.  And so on.  Thanks for asking.

dereknop:
> Yes, because there are different numbers of probes in each bin. For example in AF, the first bin contains probes with midpoint less than 0.20, and there are 10 such probes in the full list of AF probes. 

Ah that makes sense, thanks.

dereknop:
Okay, I'll grab a fresh backup and give it another go. Mine is from a few days ago.

PR #173, opened 2025-09-29
ITM-1139:  Populate database with RQ2 data for 4D experiment

nextcen-dgemoets:
~~I am not able to hit AWS ADEPT due to Zscaler, so I cannot generate alignment averages.  Brian was able to, so maybe you will as well.~~   I am now able (slowly) to generate alignment averages.

You need a recent copy of the DB for this, anything from Monday 29-Sept or later should be fine.

Please note that when we did 2D six months ago, aligned and baseline were different scripts which added rows to the multi kdma database table, and I imagine the dashboard combined the data into the form we see now in RQ2.  For this effort, I combined aligned and baseline into one script so that the dashboard should have everything it needs for RQ2 straight from the database. 
 
This script can be run with the deployment script, but I'd recommend just running it directly for now so your DB version doesn't get updated:
- `python scripts/_1_0_1_setup_p24d.py`

This script should create the `multiKdmaData4D` collection with everything needed for RQ2.  See the attached Excel file for what the fields/columns are.
[Template for 4D RQ2.xlsx](https://github.com/user-attachments/files/22602427/Template.for.4D.RQ2.xlsx)

One other quick note:  I'm not getting baseline data from one of the PIDs.  ~~I'm investigating.~~  This seems to happen when multiple PIDs have the same KDMAs.  ~~Will push a fix.~~ I have pushed a fix.

### DIFF
--- a/scripts/_1_0_1_setup_p24d.py
+++ b/scripts/_1_0_1_setup_p24d.py
@@ -5,6 +5,8 @@ from decouple import config
 EVAL_NUM = 11
 ADEPT_URL = config("ADEPT_URL")
 HIT_TA1_SERVER = True # Useful if you can't reach the TA1 server
+PYTHON_CMD = 'python3'
+FOUR_D_COLLECTION = 'multiKdmaData4D'
 
 def get_kdma_value(kdmas, kdma_name):
     '''
@@ -17,7 +19,7 @@ def get_kdma_value(kdmas, kdma_name):
 
 def main(mongo_db):
     # Create a new DB to store all the data.
-    multi_kdmas_4d = mongo_db['multiKdmaData4D']
+    multi_kdmas_4d = mongo_db[FOUR_D_COLLECTION]
     multi_kdmas_4d.drop()
 
     adm_collection = mongo_db['admTargetRuns']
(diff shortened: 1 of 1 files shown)


NeilDaniel07 (commented):
When I ran the script using the given terminal command, I received the error `FATAL: Could not find match for AF=0.3363534511058407, MF=0.13895667750000001, PS=0.5622837715384615, SS=0.9201000005576923` after `Analyzing ADM group 52 of 61 - get comparison (text|adm)`. I couldn't find this particular combination of kdma values in any row of my `text_kdmas.csv` file. Tracing the logic of the script, it's difficult to see where this combination of kdmas might be coming from (since the script just seems to be reading in the file and parsing each row directly). I'm not sure what I'm missing.

brianpippin:
> When I ran the script using the given terminal command, I received the error `FATAL: Could not find match for AF=0.3363534511058407, MF=0.13895667750000001, PS=0.5622837715384615, SS=0.9201000005576923` after `Analyzing ADM group 52 of 61 - get comparison (text|adm)`. I couldn't find this particular combination of kdma values in any row of my `text_kdmas.csv` file. Tracing the logic of the script, it's difficult to see where this combination of kdmas might be coming from (since the script just seems to be reading in the file and parsing each row directly). I'm not sure what I'm missing.

Do you have the latest DB from production?

nextcen-dgemoets:
> Do you have the latest DB from production?

Added a note to the PR about this.  Sorry, @NeilDaniel07 .

NeilDaniel07:
> Do you have the latest DB from production?

Yes, I just updated my DB yesterday using the latest backup. I also double checked the `itm-version` collection and it's `1.0.0`, which I believe is up to date. Should I also rerun the script in #167 (`python dev_scripts/get_text_kdmas.py`) to update the CSV file before running this script?

nextcen-dgemoets:
> The script in this ticket re-runs `get_p2_4d_text_kdmas.py` (the phase 2 equivalent of the script you noted, which has since been renamed).

It occurs to me that the script calls `python3` since that is what production wants.  On my Windows system, I have to change it to `python`.  If that's the case, then `get_p2_4d_text_kdmas.py` would not have re-run.

PR #210, opened 2026-03-27
ITM-1311: Add Oracle alignment to RQ2 (admTargetRuns)

nextcen-dgemoets:
<h3> NOTE:  You need the latest TA1 server (v5.6.1) released late 26-Mar-2026 to review this PR. </h3>

This script serves as a postscript for scripts 124 and 129 to add a field to RQ2 for the oracle alignment.  It reads in the probe set csv file, creates a session for each probe set / alignment target combination (always choosing choice A since it's ignored by the API anyway), then asks TA1 for the oracle alignment score.

It adds a column to the synthetic adm runs (in `admTargetRuns`) that contains the oracle alignment for that run, serving as a "ceiling" alignment for that row.

You may wish to run without TA1 connection initially (set `HIT_TA1_SERVER` to `False`), and/or use the `-c 1` switch to run a single set against the TA1 server.

If you want to run the whole thing, then you can use `python deployment_script.py`.


dereknop (changes requested):

dereknop:
I don't think this will work for anyone using their Docker setup. After adding `COPY Oracle.R .` to their Dockerfile, it seems to work for me. It also works for me using the Python setup.

dereknop (approved):
The issue is either with my local setup or the ADEPT server Dockerfile. Script looks good.

Garycheng92 (approved):
LGTM I was able to verify the new oracle_alignment values in my local db

PR #214, opened 2026-04-20
quick fix for 202604162

dereknop:
Script repairs `combinedMostLeastAligned` for an edge case participant


brianpippin (approved):
Ran for me.

PR #215, opened 2026-04-21
Itm 1321: Generate Comparison Scores for April RQ1

dereknop:
Runs the Oracle ADMs through the TA1 server. Also repopulated the MF-PS and AF-PS ADM runs so they are available for comparison scores. Generates comparison scores between participants and the observed ADMs. 

Fixes the issue of the participant who gave us the wrong PID. 

`python deployment_script.py` or `python redo.py 135`

[Dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/459)

Note: The repopulation script is not compatible with this evaluation. I am working on an update, but in the meantime, you can point at production.

### DIFF
--- a/scripts/_0_8_3_June_Collab_Comparison_Generation.py
+++ b/scripts/_0_8_3_June_Collab_Comparison_Generation.py
@@ -59,7 +59,7 @@ def main(mongoDB, EVAL_NUMBER=8):
                 elif scenario_attribute is None or scenario_attribute not in page_scenario:
                     continue
 
-                if EVAL_NUMBER != 10:
+                if EVAL_NUMBER != 10 and EVAL_NUMBER != 16:
                     adm = db_utils.find_adm_from_medic(EVAL_NUMBER, medic_collection, adm_collection, page, page_scenario, survey)
                     if adm is None:
                         continue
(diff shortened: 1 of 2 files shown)


nextcen-dgemoets (commented):
The table and download looked good.  Derek held my hand through comparing values in the table to values in the DB, and everything looked good-- except one anomaly where for some reason most/all of my Oracle alignments were `NaN` in my database.  His values are fine.  This should only happen if there's a problem with the oracle session ids, so as long as things look good for the next reviewer, we'll say it's fine and I'll go back and approve this retroactively.

Also, there were some formatting issues (e.g., crazy row heights) with the downloaded variable definitions, so I sent Derek a fixed version.

[Edit:  this comment was supposed to go on the dashboard PR for this ticket.  Oh well.]

patmole99 (approved):
Wasn't getting the Oracle alignments, but then I got a fresh DB backup and started getting all the scores I wanted to see. Derek said it was a result of running the repop script locally for my probe matcher, before testing this script. LGTM!

dereknop:
> except one anomaly where for some reason most/all of my Oracle alignments were `NaN` in my database. His values are fine. This should only happen if there's a problem with the oracle session ids, 

Unfortunately, my next reviewer had the same issue. When reviewing Patrick's code review, the `combinedSessionId` for your participants were overwritten when running the repopulation script (pointed at your local ADEPT Servers). The Oracle comparison scores use this session id which is what caused the NaN scores. Confirmed with Patrick, and we fixed it with a fresh db. This will not be a problem on production.

nextcen-dgemoets (approved):
The `NaN` issue was resolved / explained, and the updated definitions were uploaded.

PR #216, opened 2026-04-22
Triage performance calculation update

patmole99:
# PR: Update Triage Performance + Fix Alignment KDMA Fetching

## Summary

This PR includes the following updates:

- Updates **Triage Performance** calculation to account for **supplemental procedures** (aligned with June/April logic)
- Ensures **alignment KDMAs (MF/AF)** are properly fetched by restoring valid ADEPT session IDs when needed

---

### 1. Triage Performance Bug
The February probe matcher previously:
- Only considered **required treatments**
- Ignored **supplemental procedures**, even though those metrics were computed

This resulted in **underreported performance values**.

This PR fixes that by:
- Incorporating supplemental procedures into the calculation
- Matching the logic used in June and April matchers

---

### 2. Alignment KDMA Issues (NaN values)

Observed on prod:
- `MF Alignment_Desert`
- `AF Alignment_Desert`
- `MF Alignment_Urban`
- `AF Alignment_Urban`

These values were returning **NaN**, which typically indicates:
- Missing or invalid ADEPT session IDs

Fix approach:
- Rerun probe matcher and/or
- Use `ph2_repop.py` to regenerate ADEPT sessions and repopulate probe responses

---

### 1. Setup ADEPT Locally

Run a local ADEPT server and ensure sessions are populated:

```bash
python ph2_repop.py
```

> This recreates ADEPT sessions and re-sends probe responses.

---

### 2. Capture Baseline Data

- Open the dashboard (local or prod)
- Download **RQ8 table data**
- Keep this for comparison

---

### 3. Run Updated Probe Matcher

```bash
python feb2026_probe_matcher.py -i ph2_feb2026_sim_files
```

> Do **not** pass `--no_kdmas` since alignment KDMAs need to be recomputed.

Recommended:
- Let several sim files process
- Note the PIDs of processed runs

---

### 4. Validate in MongoDB

Using Mongo Compass:

Check collection:
```
humanSimulator
```

Navigate to the `actionAnalysis` section of the document for each processed run.

Verify:

#### ✅ Triage Performance Updated
- `Desert Triage Performance` OR `Urban Triage Performance`
- Should be **higher than before** when supplemental procedures are present

#### ✅ Alignment KDMAs Fixed
- `MF Alignment_Desert` / `AF Alignment_Desert`
- `MF Alignment_Urban` / `AF Alignment_Urban`
- Should **NOT be NaN**

---

## Expected Results

- Triage Performance reflects **supplemental + required treatments**
- Alignment values are properly populated (no NaNs)
- No regressions in other extracted fields

---

## Notes

- `ph2_repop.py` is **safe to rerun**, but:
  - It creates **new ADEPT sessions**
  - It is not strictly idempotent
- This PR does **not** change probe matching logic, only:
  - Triage calculation
  - Alignment reliability

### DIFF
--- a/feb2026_probe_matcher.py
+++ b/feb2026_probe_matcher.py
@@ -1439,9 +1439,24 @@ class ProbeMatcher:
         # score = hits / (tools_applied + misses) * 100
         # -------------------------
         def get_triage_performance():
+            """
+            June-style supplemental-aware triage performance.
+
+            Numerator:
+              - completed required injuries
+              - approved supplemental tools, but only once a patient has no
+                required injuries remaining
+
+            Denominator:
+              - all TOOL_APPLIED events except Pulse Oximeter
+              - plus remaining required injuries (misses)
+            """
             remaining = {p: set(inj_list) for p, inj_list in required_injuries.items()}
-            correct_completed = 0
+            supplemental = copy.deepcopy(SUPPLEMENTAL_PROCEDURES.get(env.lower(), {}))
+            supplemental_tracker = {}
+            correct_tools_applied = 0
             total_tools_applied = 0
+            just_completed = None
 
             for row in data:
                 ev = _get_cell(row, idx, "EventName")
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  I followed the instructions of the PR.  I verified the updated triage performance and non-NaN KDMAs of a processed PID.  I loaded a pid that wasn't yet processed and observed NaN kdmas and triage performance that matched the downloaded file from prod.  Once the probe matcher processed that PID, I refreshed mongo and observed a higher performance measure, and KDMAs that were no longer non-NaN.  My naïve perusal of the code revealed no issues to me.

Garycheng92 (approved):
Looks good, I was able to see the increase in performance values and KDMAs no longer have NaN

PR #217, opened 2026-04-23
ITM 1246

ktabasco-prog:
**Bug Issue:**
 
The surveyResults and userScenarioResults MongoDB collections contain a significant amount of invalid and outdated data, which has led to reduced data quality. This includes:

- Documents with missing or null participant IDs
- Documents with participant IDs not present in the participantLog collection
- Documents containing obvious test data

 
**Fix:**
 
This PR introduces a cleanup script that validates participantID values across both collections and removes invalid records based on the following criteria:

- Documents where the participant ID is missing or null
- Documents where the participant ID does not exist in participantLog, only when evalNumber >= 4
- Documents containing obvious test data, regardless of evalNumber

This approach ensures that only reliable, production-relevant data is retained while preserving earlier data (pre-DRE).
 
**Testing Instructions:**

1. Check out the ITM-1246 branch in the ingest repository.

- Run the script with **DELETE = False** --> **This performs a dry run**
- Verify that document counts match: The script output vs. MongoDB queries (via Compass or command line)
- Cross-reference the printed _id values with documents in MongoDB to confirm correctness

2. Once validated, update **DELETE = TRUE**

- Re-run the script and confirm:
- The reported deletion counts match expectations

**Expected Output:**
 
**Dry Run**
 
No documents should be deleted, and these counts should be printed:
 
- SurveyResults | NULL PIDs Total Documents: 41
- SurveyResults | INVALID PIDs Total Documents: 180

- UserScenarioResults | NULL PIDs Total Documents: 0
- UserScenarioResults | INVALID PIDs Total Documents: 38

- **SurveyResults | OLD PIDs BEFORE 4 Total Documents: 64**

 
**Delete Run**
All **259 documents** should be deleted and the script should print the same counts as the breakdown
 
**Thoughts & Comments:**
 
After running the script, there are still some pre-DRE (evalNumber < 4) records in surveyResults that appear to be test data. Examples include participant IDs such as "1234", "12341234", "not real", or other random character strings.
 
At this stage, there is no reliable validation source (like participantLog) for these earlier records, making it difficult to confidently distinguish invalid data from legitimate entries.
 
I am very much open to any ideas and suggestions for this particular issue.

**UPDATE:** I added an extra function to catch all junk data from old documents in surveyResults that didn't match valid PID syntax. Please re-follow the steps above to ensure it works and has the expected output! **The updated deleted document total is 323.**


dereknop:
> After running the script, there are still some pre-DRE (evalNumber < 4) records in surveyResults that appear to be test data. Examples include participant IDs such as "1234", "12341234", "not real", or other random character strings.

I can take a closer look at this, but off the top of my head, I think we have always had the format of participant IDs starting with the year. So any valid PID should start with `2024`, `2025`, or `2026`.

patmole99 (approved):
The script ran successfully for me. My logged numbers matched your logged numbers on the first delete run and then I got 0 for each when running it a second time, so it seem like they were deleted successfully. I would switch `DELETE` to be `True` before merging, but it LGTM!

ktabasco-prog:
> > After running the script, there are still some pre-DRE (evalNumber < 4) records in surveyResults that appear to be test data. Examples include participant IDs such as "1234", "12341234", "not real", or other random character strings.
> 
> I can take a closer look at this, but off the top of my head, I think we have always had the format of participant IDs starting with the year. So any valid PID should start with `2024`, `2025`, or `2026`.

I added another query to sift through those old documents pre DRE to catch all that junk data. Thanks for the suggestion, let me know what you think and if it works for you! It looks like everything is working on my end

dereknop (approved):

PR #220, opened 2026-05-05
ITM-1334: Adds KDMA calculations using other subpopulation group

dereknop:
This script will calculate KDMA scores for participants using the subpopulation group they were NOT a part of. We already have the scores using the subpopulation group they were a part of. I have updated my open [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/464) to handle these new fields. 

`python deployment_script.py` or `python redo.py 137`

Note: you'll want your `.env` pointed at production, not local.

After running, the eval 16 results in `userScenarioResults` should now have a `otherSubKDMA` field containing the scores we got back from the ADEPT server.

### DIFF
--- a/scripts/_1_3_7_kdma_other_subpop.py
+++ b/scripts/_1_3_7_kdma_other_subpop.py
@@ -1,0 +1,55 @@
+import requests
+from decouple import config
+from bson import ObjectId
+ADEPT = config('ADEPT_URL')
+
+# spurious text documents ids that need to be deleted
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.

patmole99 (approved):
I do see the `otherSubKDMA` field now in the `userScenarioResults` documents.

PR #222, opened 2026-05-06
ITM-1348: Open World ADM Cleanup

patmole99:
## Summary

This is just a short script to clean up the database and remove some unnecessary ADM runs.

## How To Test

First run this command and confirm that there are 55 documents that will be removed at the top of the log:

`python redo.py 139 false true`

Once you confirm that 55 documents will be removed run it again without the Boolean flags. This will remove the documents:

`python redo.py 139`

Then run it once more to confirm that they were successfully removed and the script finds no more documents.

### DIFF
--- a/scripts/_1_3_9_remove_unwanted_adm_runs.py
+++ b/scripts/_1_3_9_remove_unwanted_adm_runs.py
@@ -1,0 +1,102 @@
+"""
+ITM-1338 / Open World ADM cleanup
+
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  It deleted the two ADMs, and repeated calls didn't make any further changes.
The script is arguably overkill, but I've been known to have overkill tendencies myself.

dereknop (approved):
👍

PR #223, opened 2026-05-07
ITM-1332: Quick fix to switch to upload to results

patmole99:
This switched the rescore in mongo db to upload to `results`, not `evaluation.results`.

Rerun `py redo.py 138 true true true` and ensure the values in results were updated.

### DIFF
--- a/scripts/_1_3_8_open_world_rescoring.py
+++ b/scripts/_1_3_8_open_world_rescoring.py
@@ -1,19 +1,19 @@
 """
 ITM-1332 / ITM-1337: Rescore Feb2026 Open World desert ADM target runs.
 
-This script recalculates TA1/ADEPT session alignment results for the Feb2026
-Open World desert ADM runs identified, then writes the recalculated result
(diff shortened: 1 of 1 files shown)


dereknop (approved):

scripts/_1_3_8_open_world_rescoring.py
```diff
@@ -3,16 +3,16 @@
 
 This script recalculates TA1/ADEPT session alignment results for the Feb2026
 Open World desert ADM runs Darren identified, then writes the recalculated result
-object back to admTargetRuns.evaluation.results.
+object back to admTargetRuns.results.
 
 Examples:
-  python redo.py 137 false true false
+  python redo.py 138 false true false
       Dry-run, verbose, do not enforce the expected 16+16 document count.
```
nextcen-dgemoets: While we're here, there are now four true/falses, not the three documented here.
Also, remove my name.  It can say "recalculates TA1/ADEPT session alignment results for the final Feb2026
Open World desert ADM runs, then writes the recalculated result [...]"

PR #224, opened 2026-05-07
ITM-1349:  Fix April OW yamls

nextcen-dgemoets:
Fixed scenario/probe ids, and medical attribute of broken arm patient.


patmole99 (approved):
Added these yaml files to my local ADEPT server, rebuilt it, ran the April probe matcher and was able to get KDMA values.

dereknop (approved):

PR #225, opened 2026-05-15
ITM 1201: Expected KDMA TA1 Mismatch Cleanup and Future Occurrence Detection

NeilDaniel07:
**Original Ticket:** [Here](https://nextcentury.atlassian.net/browse/ITM-1201?atlOrigin=eyJpIjoiNWFmMTAwZmYxOTIwNGI5NWFmZmQyNGMwZDNkOTQ0MDgiLCJwIjoiaiJ9)

## Description

This PR cleans up KDMA result-set contamination observed in ADEPT June 2025 `admTargetRuns` documents in the database originally flagged by @dereknop when it appeared in the dashboard. It also adds a reusable detector to help identify future recurrence of the same class of issues to easily detect and resolve them, minimizing downstream impact.  The original issue was that some completed single-KDMA ADEPT evaluation documents contained extra entries in `results.kdmas`. For example, a document targeting a single KDMA such as `ADEPT-June2025-merit-0.6` could contain the expected `merit` KDMA along with unrelated KDMA entries from another single-attribute run. These extra entries were not expected for single-KDMA targets and appeared to originate from an incorrect computed KDMA profile data returned from the TA1 API rather than from anything in our own codebase. I investigated this by reviewing the affected Mongo documents, inspecting the TA3 server path that writes `results.kdmas`, and building a standalone diagnostic replay script that exercised TA1 directly using recorded probe responses from the affected documents. The replay script called the relevant TA1 endpoints used during evaluation (`new_session`, `response`, `get_ordered_alignment`, `alignment/session`, and `computed_kdma_profile`) and tested full replay of an ADM run with randomized document ordering, mid-probe checks, session-end checks, and concurrent execution with 8 threads. I was not able to reproduce the contamination locally. After discussion with @nextcen-dgemoets, our final decision was to clean the known bad data and include a reusable detector for future evaluation data.

This PR adds two files, a cleanup script at `scripts/_1_4_0_clean_adept_june2025_single_kdmas.py` and a recurrence detector at `dev_scripts/check_kdma_result_sets.py`. The cleanup script offers a one-time resolution for the contaminated ADEPT June 2025 single-KDMA target documents in `dashboard.admTargetRuns`. It infers the expected KDMA from `evaluation.alignment_target_id`, filters `results.kdmas` to retain only entries whose `kdma` matches the expected value, and updates only the `results.kdmas` field. It intentionally does not modify `history`, `evaluation`, `results.alignment_score`, `results.ta1_session_id`, or any other document fields.

The recurrence detector is intentionally not a numbered script under `scripts/` run only once as the probe matcher updates a database version. Instead, `dev_scripts/check_kdma_result_sets.py` is a reusable validator that can be run manually after future evaluations. It detects persisted documents where the KDMA set expected based on the ADEPT target ID does not match the KDMA set stored in `results.kdmas`. Unlike the cleanup script, the detector is not hardcoded to `ADEPT-June2025`. It supports June 2025 full-name targets and Feb2026 alias targets in Mongo, including formats such as:

- `ADEPT-June2025-affiliation-0.5`
- `ADEPT-June2025-affiliation_merit-0.0_1.0`
- `Feb2026-AF-1`
- `Feb2026-AF1-PS1`
- `Feb2026-MF2-SS8`

It supports `affiliation`, `merit`, `search`, and `personal_safety`, along with aliases `AF`, `MF`, `SS`, and `PS`. Multi-KDMA targets are parsed as multi-KDMA expected result sets rather than automatically being treated as contamination.

## Detector CLI Flags
`--mongo-url URI:` MongoDB connection URI to use for the read-only detector. If omitted, the script uses its default/local Mongo connection behavior.
`--db NAME:` Mongo database name to scan. Defaults to `dashboard`.
`--collection NAME:` Mongo collection name to scan. Defaults to `admTargetRuns`.
`--target-prefix PREFIX:` Restrict validation to target IDs beginning with the provided prefix, such as `ADEPT-June2025` or `Feb2026`. If omitted, the detector scans all parseable documents in the collection.
`--domain DOMAIN:` Restrict validation to documents matching a specific evaluation domain, if domain metadata is present.
`--kdma NAME:` Add or override a supported KDMA name for target parsing. Can be used if future evaluations introduce additional KDMA names beyond the defaults.
`--alias ALIAS=KDMA:` Add or override a target-ID alias mapping, such as `AF=affiliation` or `MF=merit`. Useful if future target IDs introduce new shorthand labels.
`--include-not-requested:` Include documents where `alignment_score == "Not requested"` in validation. These are skipped by default to avoid expected false positives.
`--include-open-world:` Include open-world/full-profile documents in validation. These are skipped by default because they may correctly contain broader KDMA profiles than the set implied by the target alone.

## Disclaimers

1. Since I was not able to reproduce the original contamination locally, this PR does not isolate or fix the original TA1-side root cause.
2. The detector skips open-world ADM run results and `alignment_score == "Not requested"` documents by default. This avoids false positives in Feb2026 data where those cases legitimately contain a full computed profile or no requested KDMA result (respectively). Running with `--include-not-requested` and `--include-open-world` can override these assumptions if a developer would prefer this option. 
3. No TA3 server runtime guard is included here. A server-side warning might catch recurrence closer to the moment it occurs (since this is the point of our interaction with TA1), but that would change runtime evaluation behavior and likely belongs in a separate PR if desired.
4. The detector relies on parseable target IDs. Future KDMA names, aliases, or different target formats may require `--kdma` and `--alias` flags to be set, or updates to this checked-in code.

## Investigation Artifacts

Feel free to download  [ITM-1201.zip](https://github.com/user-attachments/files/27789421/ITM-1201.zip). It contains a text version of nine contaminated documents (`InfectedDocumentOne.txt` through `InfectedDocumentNine.txt`) and the standalone TA1 replay script I developed during my investigation of the bug. These files aren't committed and not required to run the cleanup or recurrence detector. They are useful if reviewers want to inspect the original affected document shape, compare `evaluation.alignment_target_id` against `results.kdmas`, or review how I replayed calls to the TA1 API just as our evaluation server produces them when supporting an ADM run. 

## How to Test

1. Run the cleanup script using `python redo.py 141`. You can also run it with lower-verbosity if you prefer using `python redo.py 141 false`. You should see that 14 documents were identified as contaminated and cleaned. By default, the script will log the details of these documents, so you can verify that their `results.kdmas` state has been resolved in Mongo Compass once the script execution is complete.
2. Rerun the cleanup script once again and confirm idempotency. After the first successful run, the second run should report no remaining documents requiring cleanup. 
3. Run the detector against Feb2026 data using default skip behavior using `python dev_scripts/check_kdma_result_sets.py --target-prefix Feb2026`. The expected result based on the current database state should look something like this:

```text
matched=51665
suspicious=0
skipped_not_requested=1927
skipped_open_world=128
```
4. Run the broader diagnostic mode that does not ignore open world ADM run mismatches and documents where KDMAs were explicitly not requested. `python dev_scripts/check_kdma_result_sets.py --target-prefix Feb2026 --include-not-requested --include-open-world`. You should now see a value close to  `suspicious=2055`.  This verifies that the detector can identify mismatches when diagnostic inclusions are enabled, while default behavior avoids false positives.
5. Run the detector against ADEPT June 2025 after cleanup. You should see `suspicious=0` once again. You can use the command `python dev_scripts/check_kdma_result_sets.py --target-prefix ADEPT-June2025` for this.
6. Run the detector on all documents in the `admTargetRuns` collection using `python dev_scripts/check_kdma_result_sets.py`. Once again, no documents should be flagged as contaminated.

**NOTE:** I also created synthetic data that covered clean single-KDMA targets, clean multi-KDMA targets, extra KDMAs, missing KDMAs, not-requested skips, open-world skips, and opt-in diagnostic inclusion for unit testing during development. If you would like to verify the script's behavior with these cases or add edge cases of your own, feel free to use this Python script as a foundation: [ITM-1201 Unit Testing.py](https://github.com/user-attachments/files/27789941/ITM-1201.Unit.Testing.py). Let me know if you would like any more details (please ask via GitHub as opposed to Teams) and I would be happy to provide them.


patmole99 (approved):
Tested and reviewed the PR. The cleanup script successfully cleaned the 14 contaminated ADEPT June 2025 documents, and a second run confirmed idempotency with 0 remaining updates.

I also verified the cleanup only updates `results.kdmas`, and that the detector lives under `dev_scripts/` as a reusable manual validation script rather than a numbered migration.

The detector returned the expected results for Feb2026, ADEPT-June2025, and the full collection. Everything looks good to me.

nextcen-dgemoets (approved):
LGTM.  The scripts ran as expected and we'll be able to use this moving forward to detect this anomaly.

After the June2026 observation runs are complete, I'll run:
- `python dev_scripts/check_kdma_result_sets.py --target-prefix ADEPT-June2026`

to see if we have any more anomalies.

PR #229, opened 2026-06-05
Regen Del materials

dereknop:
I used an outdated template when I initially created the delegation materials. This led to two updated questions being omitted. This will regenerate the materials.

`python deployment_script.py` or `python redo.py 143`

If you go to http://localhost:3000/review-delegation you should see these questions when u click on a medic:
Screenshot 2026-06-05 at 11 47 58 AM

Only one of these will ever be presented at a time in the actual delegation survey. 

Check out [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/480)


nextcen-dgemoets (approved):
> Only one of these will ever be presented at a time in the actual delegation survey.

Yes, that's what I observed.

joshuale-caci (approved):

PR #230, opened 2026-06-09
ITM-1373: Ingest Observation YAMLs June 2026

dereknop:
For the ADM results pages to work, we need to ingest the observation YAML files. 

`python deployment_script.py` or `python redo.py 144`

Now we can go to [admin](http://localhost:3000/admin) and toggle on the adm results pages
Screenshot 2026-06-09 at 10 23 13 AM

Each of the three pages should now populate for June 2026 (page may need a refresh first).
http://localhost:3000/adm-results
http://localhost:3000/results
http://localhost:3000/adm-probe-responses

### DIFF
--- a/scripts/_1_4_4_june_scenario_files.py
+++ b/scripts/_1_4_4_june_scenario_files.py
@@ -1,0 +1,29 @@
+import os, yaml
+
+    
+SCENARIOS_FOLDER = 'phase2/june2026/admrun'
+
+
+def main(mongo_db):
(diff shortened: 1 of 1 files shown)


patmole99 (approved):
I noticed the forever "Loading..." on the results page and some empty columns in the data, but it looks like there are new tickets to address these in the future and everything else LGTM.

Garycheng92 (approved):
LGTM! Pages populate for June 2026

PR #232, opened 2026-06-12
ITM-1383: ADM Cleanup June 2026

dereknop:
This script fixes a bug in Kitware's ADM runs where one set of ADM runs on AF, PS, and AF-SS were named incorrectly. It also cleans up the ADM names (strips the UUID and adds the date of the run to the end of the name). Lastly, it removes the ADM runs that had 'Random' in their name since these were never supposed to be added to our DB. 

`python deployment_script.py` or `python redo.py 145`

An easy way to check if this ran correctly is using http://localhost:3000/adm-results

On prod:
Screenshot 2026-06-12 at 12 29 41 PM

Local:
Screenshot 2026-06-12 at 12 30 09 PM

You'll notice on prod there are two instances of the Direct Regression DeepSeek model. After running, we have the names cleaned up, and we now have a Direct Regression Mistral model.

### DIFF
--- a/scripts/_1_4_5_adm_cleanup_june2026.py
+++ b/scripts/_1_4_5_adm_cleanup_june2026.py
@@ -1,0 +1,54 @@
+from datetime import datetime
+def main(mongo_db):
+    adm_runs = mongo_db['admTargetRuns']
+    bad_boys = [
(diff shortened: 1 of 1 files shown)


joshuale-caci (approved):
👍 UUIDs were stripped and the bad boys are gone!

Garycheng92 (approved):

nextcen-dgemoets (approved):
LGTM!
I know you had two reviewers, but I just really wanted to review something that required a working Docker environment...

PR #235, opened 2026-06-25
ITM-1359 Add new metric estimated_evac_decision_time to TCCC data

Garycheng92:
Adds `Estimated Evac Decision Time` + `Estimated Evac Decision Time Confidence`

TCCC files from before June 2026 don't log when the participant got asked who to evacuate, so we can't directly measure how long that decision took. This PR backfills an estimate for those older files and writes it onto each participant's `tcccResults` doc. It's a one-time thing — the June files log this directly through the TCCC lambda function in S3, so they're not touched here.

The main pieces:

`misc_scripts/evac_response_window.py` — the calculation
`scripts/_1_4_8_estimated_evac_decision_times.py` — the migration that writes the metrics to Mongo
`misc_scripts/llm_picks.json` — the committed LLM picks the calculation reads
What we're measuring is the gap between when the participant finished triaging and when their evac answer got logged. Figuring out that start point happens one of two ways.

If they ran the full scenario (10:45) and were still treating when time ran out, the evac prompt was forced by the timer at that mark — so we actually know the start exactly and mark it HIGH confidence.

If they finished early, we estimate the start from three signals. The first two come from the participant's voice transcript, but only the last 90 seconds before they answered — that's the window where the wrap-up and evac talk actually happens:

- Heuristic — a keyword scan of that 90-second window for evac language ("evac", "this one", "finished", etc.), weighted by how strongly each phrase points to someone actually picking a patient.

- LLM — an LLM pass that read the same window and judged where they started deciding. Those picks live in misc_scripts/llm_picks.json (a committed file, generated once, not a live API call).

- Last treatment — the timestamp of their last real treatment in the scenario (blankets, lollipops and other comfort items don't count). This one comes from the action log, not the transcript.

To turn those into one start time, we take the earliest of the two voice signals as the moment they wrapped up — unless their last treatment came after that, which means they said they were done but kept working, so we use the treatment time instead. If there's no voice to go on at all, we fall back to the last treatment. The decision time is then just that start subtracted from when their answer was logged.

Confidence is how tightly the three signals agree: within 5 seconds is `HIGH`, within 30 is `MEDIUM`, anything wider — or only one signal to go on — is `LOW`.

Some participants make their evac picks silently — clicking through the interface without saying anything. With no transcript to read, the heuristic and LLM come up empty, so the estimate leans entirely on the last-treatment timestamp and the confidence lands at LOW, since there's nothing to corroborate it against. That's expected rather than a failure — it just means we can't pin those ones down as tightly.

One participant (`202604130`) has a `tcccResults` doc but won't get an evac time. Their session ended early, around the 4-minute mark, before the evac question ever came up — so there's no decision to measure, and its evac fields are left blank on purpose.

To test:

Grab the latest mongo backup from S3
Run `python redo.py 148`
Should print `Finished updating 49 documents`

The dashboard should show the new columns, `Estimated Evac Decision Time` + `Estimated Evac Decision Time Confidence`


patmole99 (approved):
After running the script I now see `Estimated Evac Decision Time` and `Estimated Evac Decision Time Confidence` for the pre June 2026 sim runs in the TCCC table.

jaudick (approved):
image

I see it. I also circled in green participant 202604130 and that it is blank like you described

dereknop (approved):

PR #236, opened 2026-07-01
Itm 1387: RQ1 and cleaning up a mess

dereknop:
WARNING: This script will probably take about 2 hours to run in full. Part 1 will take 0.5 seconds, Part 2 will take about 20 minutes, and Part 3 will take the rest. Part 3 does not need to be run in full to confirm it is working correctly (maybe for about 10 minutes or so). 

### **Part 1** 
 Spurious text scenario results documents will be deleted. This can be checked via the participant progress page in the dashboard. There should be no participant for June 2026 with more than 6 text documents.

### **Part 2** 
 ADEPT changed their alignment scoring between Washington and Minnesota... The script will rescore all of the observed ADMs used in the delegation survey. That is, the alignment score between the ADM and its alignment target. Next, we will rescore the text scenario results of the humans as needed. None of the Minnesota stuff needs to be overwritten but as you'll see, there is a lot of Washington data that is affected. The affected data should now automatically be marked as exempt in the dashboard. All of these discrepancies will be added to an Excel file `eval17_rescore_discrepancies.xlsx` (this has already been passed to Jennifer).

Note: the ADM session IDs were wiped from the ADEPT server, so Part 2 creates new session IDs for them and replaces the session IDs in our database so that the comparison scores can be generated in Part 3.  

### **Part 3**
Finally, calculate the comparison scores between the humans and the observed ADMs. Each block, except for AF-SS, will have both a Binary and a trinary comparison. You don't need to let it finish running. You will know that you are in part three when your console looks like this:
Screenshot 2026-07-01 at 12 09 08 PM

When this finishes running for me, I will post my results either here or in the dashboard channel so you can spot check with what you have run. [RQ-134 data_June2026.xlsx](https://github.com/user-attachments/files/29569311/RQ-134.data_June2026.xlsx)

`python deployment_script.py` or `python redo.py 149`

See [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/493)

### DIFF
--- a/scripts/_0_8_3_June_Collab_Comparison_Generation.py
+++ b/scripts/_0_8_3_June_Collab_Comparison_Generation.py
@@ -35,7 +35,7 @@ def main(mongoDB, EVAL_NUMBER=8):
             session_id = entry.get('individualSessionId') if is_individual_mf else entry.get('combinedSessionId')
         elif EVAL_NUMBER == 17:
             # SS-assess feeds the AF-SS 2D block (scored under its own combined AF+SS session);
-            # everything else (AF/PS, binary or trinary) uses its group's combinedSessionId.
+            # every other doc (AF/PS, binary or trinary) uses its group's combinedSessionId.
             session_id = entry.get('AF-SS_sessionId') if 'SS' in scenario_id else entry.get('combinedSessionId')
         else:
             session_id = entry.get('combinedSessionId')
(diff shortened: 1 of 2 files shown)


joshuale-caci (approved):
Looks like my values look very similar to yours and the script ran for a while I got up to 250 before quitting. Also on windows there is a spreadsheet compare app that I never knew existed. very useful!

dereknop:
> Looks like my values look very similar to yours and the script ran for a while I got up to 250 before quitting. Also on windows there is a spreadsheet compare app that I never knew existed. very useful!

After further review, Josh's scores and my own can differ by (1 x 10⁻⁷). It seems like there is some rounding difference somewhere along the way with either the ADEPT server or our own OS. Waiting to hear back from Jennifer if this is statistically significant or not.

patmole99 (approved):
Had to restore my db and point at prod adept, but after that my values matched Derek's.

PR #237, opened 2026-07-02
ITM-1398: June2026 RQ2

nextcen-dgemoets:
This is the June2026 RQ2 PR.  It combines two steps that were performed as separate PRs back in the Feb2026 eval.

<h3>Part 1:  Generate probe sets</h3>

For RQ2, ADMs were run on evaluation (hold-out) probes at a variety of alignment targets to collect probe responses.
For binary probes, we post-construct N probe sets of 32 probes (8 probes per attribute) based on Latin square selection.
For trinary probes, we post-construct N probe sets of 16 probes (8 probes per attribute, AF & PS) based on random selection.
This step just creates the probe sets, and outputs two csv files (one binary and one trinary) with the set information as input to the follow-on ingest script that will collect the ADM responses from the database and run aligned vs. baseline alignment comparison across targets and probe sets.

See [the Feb2026 ticket](https://nextcentury.atlassian.net/browse/ITM-1250) or the script documentation for the high-level algorithm.

**NOTE: The 151 script now invokes the probe set generators, but I think it's still a good idea when reviewing to break it up like in this PR.  Just note that Part 2 of this PR will replace the probe sets you generated in Part 1.**
**NOTE2: This PR has been expanded to include MF binary probes/probesets.**

To test:
- Run `python dev_scripts/gen_RQ2_probesets.py` and inspect the resulting probe set csv in `phase2/jun2026/RQ2-probesets.csv`. After this PR's script is run on prod, this csv file will be added to source control as part of a future ingest PR. Check out verbose output (-v) for better insight into the script and how the probe sets are constructed.
  - The resulting csv should contain 25 3D (32-probe) probe sets, constructed by combining the 8-probe attribute-specific probe sets (8x4 = 32). The column headers and probe IDs should indicate the attribute.
- Then run the trinary equivalent:  `python dev_scripts/gen_RQ2_trinary_probesets.py` and inspect the resulting probe set csv in `phase2/jun2026/RQ2-probesets-trinary.csv`.
  - The resulting csv should contain 25 2D (16-probe) probe sets, constructed by combining the 8-probe attribute-specific probe sets (8x2 = 16). The column headers and probe IDs should indicate the attribute.

<h3>Part 2:  Create synthetic ADM runs</h3>

This script (151) reads in the csv files you generated above, collects the ADM responses from the database, feeds the probes and responses to TA1 to collect alignment and kdma values, and creates synthetic ADM runs as if each ADM had run on each probe set at each alignment target.  It also clears out some bad ADMs and renamed some accidentally mis-named ADMs.

To test, first do a run of 25 subsets without connection to TA1:
- Make sure you have the latest database snapshot from prod (because you'll need all of the TA2 ADM runs).
  - `dashboard-mongodb-daily_20260702_201118.tar`, which will have a date of about 4:10 PM or so because it was manually run so that it has today's data.  Note that it won't have data from certain recent ingest scripts (e.g., the probe matcher).  So you might want to back up your current DB first and restore it when you're done reviewing...up to you.
- set `HIT_TA1_SERVER` in 151 to `False`
- Run `python redo.py 151`
- Inspect MongoDB Compass for good data.  You should have 336 non-synthetic adm runs and 8400 synthetic ADM runs.  You could spot check that other data looks good, but remember that the TA1 data is fake.
- set `HIT_TA1_SERVER` in 151 back to `True`
- You can point to the production TA1 server, or your local version if it's faster and has all of the June2026 scenarios and targets from previous PRs.
- So that it doesn't take forever, maybe run the script manually with an n=2, e.g.
  - `python scripts/_1_5_1_june_2026_rq2.py -c 2`
  - Or if you like waiting, `python redo.py 151`
- Look at the `admTargetRuns` collection in Mongo and filter on `evalNumber: 17` and `synthetic: true` to see the output of the script.  If you ran with `-c 2`, there should be 624 records.
  - This time, you can spot-check ensure that there are valid alignment scores and kdmas (binary: AF, PS, SS, AF-SS; trinary AF, PS).
- As always, please look at the code.


dereknop (approved):
This looks good. Checked documents in `admTargetRuns` and they look how you described/I would expect.

Image

patmole99 (approved):
The csv files looked good, the scripts ran successfully, and the ADMs were populated into my mongo db.

Garycheng92 (approved):
I see the ADMs in my mongo db and the scripts all ran without any issue

joshuale-caci (approved):
MF runs were present and no errors for me!

PR #239, opened 2026-07-16
ITM 1414: Rescore ALL of eval 16 (April 2026)

dereknop:
Note: I accidentally pushed all of my work to branch `itm-1403` even though the ticket is itm-1414.

Please grab the server release `5.6.2`. You'll then need to make sure it has the right YAML files in the `data/scenario` folder. I will attach them [here](https://github.com/user-attachments/files/30096219/yaml_files.zip).

Point your `.env` to your local server. 

This script rescores everything from eval 16. Participant text-based scenarios, observed adms, and comparison scores between the two. It will take several hours to do all of that since this branch of the server doesn't have some of the quality of life improvements that were made between evals 16 and 17. 

`python deployment_script.py` or `python redo.py 153`

You'll likely need to wait a while before checking the [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/505) when it is done running.

### DIFF
--- a/ph2_repop_eval_16.py
+++ b/ph2_repop_eval_16.py
@@ -25,8 +25,6 @@ def get_kdmas(session_id, enable_subpop=None):
 
 
 def most_least_aligned(session_id, targets, enable_subpop=None):
-    """One get_ordered_alignment call per target (targets=[None] == skipKdmaFilter).
-    Drops synthetic '-group-' rows exactly like the dashboard does."""
     responses = []
     for target in targets:
         params = {"session_id": session_id}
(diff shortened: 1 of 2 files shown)


joshuale-caci (approved):
Script ran successfully. I ran into some of these warnings but I assume it has something to do with our data not about the script. (a small number of them said no survey found)

Image

ktabasco-prog (commented):
The script ran successfully, and I spot-checked several of the records. I did notice that participant IDs 202604125 and 202604129 appeared in the output with "no survey found" messages. I couldn't find those IDs in the human-to-adm-comparison collection, although I did find them in other collections (e.g. userScenarioResults).

Am I interpreting the log output correctly, meaning that the "no survey found" messages simply means those participant ids don't have documents in certain collections? Is that the intended behavior?

image

patmole99 (approved):

PR #242, opened 2026-08-03
ITM-1426: Rescore July 2025 ADM subsets

dereknop:
Point your `.env` to your local adept server. Replace your local adept server with version 4.5.2. If you need it, I can send you a zip, but it is too large to upload here. 

`python redo.py 155` or `python deployment_script.py`

This should overwrite the stored ta1 session ID and the kdma score for the July 2025 subset runs (e.g. July2025-PS1-eval) http://localhost:3000/adm-probe-responses to check that the updated results are populating.

### DIFF
--- a/scripts/_1_5_5_june2025_check.py
+++ b/scripts/_1_5_5_june2025_check.py
@@ -1,0 +1,56 @@
+'''
+Designed to rescore JULY 2025 adm subset runs using server version 4.5.2
+'''
+import requests
+from decouple import config
+
+import utils.db_utils as db_utils
+
+ADEPT_URL = config("ADEPT_URL")
+
+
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (approved):
LGTM.  The values were updated in the dashboard.

NOTE TO REVIEWERS:
- If you're using the TA1 server from the command line, you should delete the redacted versions from the server first:
  - `rm openapi_server/data/scenario/*redact*`
- To check that the values populated, compare any of the PS subset scenarios, e.g. `July2025-PS2` to what's on prod.  All of the PS values changed.

patmole99 (approved):
The values updated successfully for me.

PR #243, opened 2026-08-10
ITM-1430: Ingest Feb2026-MF-PS2 yaml file

dereknop:
This YAML file needs to be added to the `scenarios` collection for these adm runs to populate on the adm probe responses page. ADEPT would like to review these adm runs. 

`python deployment_script.py` or `python redo.py 156`

Then go check out http://localhost:3000/adm-probe-responses. If you click on Feb 2026, you should now see `Feb2026-adept-observe-MF-PS2` as an option. Clicking on it should populate a baseline and an aligned table.


brianpippin (approved):
Worked for me!

nextcen-dgemoets (approved):
After running the script, the MF-PS2 scenario was in the probe responses.

PR #244, opened 2026-08-12
ITM-1428: remove bad adm runs

dereknop:
The adm probe matcher already ran on these new runs, so all we are doing is deleting the adm runs that we were told are invalid. 

`python deployment_script.py` or `python redo.py 157`

Should delete 22 bad adm runs

[Dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/516)


nextcen-dgemoets (approved):
Before running the script (on yesterday's prod backup db), there were adm runs from all three identified "`BAD_ADMS`".  After running the script, they were gone, and the script reported, `Deleted 22 bad adms`.

nextcen-dgemoets (approved):
> We tend to refer to it as Zeroshot Comparative Regression, not the other way around. [...]

Based on Emily's response, maybe the Part 2 "ComparativeRegression-Zeroshot" ADMs  should be renamed "Zeroshot-ComparativeRegression" (like the Part 1 one), but it's not critical and doesn't prevent me from approving.

Garycheng92 (approved):
Adm runs were removed after running the script

PR #245, opened 2026-08-19
ITM-1432 Derive probe ids from admTargetRuns

Garycheng92:
This PR backfills every document in the `admTargetRuns` collection in order to speed up dashboard load times for the RQ2 table. On the dashboard, each time the query fires for RQ2, theres a fallback where a top level list of probe ids is created for each document in the collection which is a series of nested loops that reads the `history` array which I found were the cause of the slow load times. This script offloads the backfill from the query and will run as a weekly jenkins job to backfill new documents added to the `admTargetRuns` collection. There will be a follow on [dashboard PR](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/518)

To test:

- Grab mongo backup from S3
- Run `python dev_scripts/derive_probes.py`
- You should see `68120 36` Where 68120 are the documents that have no probe ids backfilled and 36 skipped documents that have no history array.
- Run `python dev_scripts/derive_probes.py true`
- You should see `0 36` where 0 means all documents with a history array now has a top level array of probe ids

Screenshot 2026-08-19 154340


misc_scripts/derive_probes.py
```diff
@@ -0,0 +1,28 @@
...
+
+def main(mongo_db, write_to_db=False):
+    updated = 0
+    skipped = 0
+    adm_collection = mongo_db['admTargetRuns']
+    missing_filter = {'$or': [{'probe_ids': {'$exists': False}}, {'probe_ids': []}]}
+    for doc in adm_collection.find(missing_filter):
+        ids = derive_probes(doc)
+        if not ids:
+            print(doc['_id'])
```
dereknop: I'd remove this print

dereknop (commented):
Works fine for me. No issues with the script itself. I do think that `misc_scripts` and `dev_scripts` should just be combined into one folder (idc which) since they pretty much are serving the same purpose. 

Also, I think a one line script should be added to `scripts` that calls this. That way when we deploy, this will automtically run.

misc_scripts/derive_probes.py
```diff
@@ -0,0 +1,28 @@
+from pymongo import MongoClient
+from decouple import config
+
+def derive_probes(doc):
+    if doc.get('probes'):
+        return [p['probe_id'] for p in doc['probes']]
+    return [h['parameters']['probe_id'] for h in doc.get('history', []) if h['command'] == 'Respond to TA1 Probe' and 'probe_id' in h['parameters']]
+
+def main(mongo_db, write_to_db=False):
```
dereknop: default this field to True for when we run on prod

dev_scripts/derive_probes.py
```diff
@@ -0,0 +1,28 @@
+from pymongo import MongoClient
+from decouple import config
+
+def derive_probes(doc):
+    if doc.get('probes'):
+        return [p['probe_id'] for p in doc['probes']]
+    return [h['parameters']['probe_id'] for h in doc.get('history', []) if h['command'] == 'Respond to TA1 Probe' and 'probe_id' in h['parameters']]
+
+def main(mongo_db, write_to_db=False):
```
Garycheng92: I'll swap it in after this PR is approved

dereknop (approved):

brianpippin (approved):
No issues running on my end.

PR #246, opened 2026-08-20
Itm 1438: adm rescore w fixed yaml files

dereknop:
Take the updated openworld files from this pr and drop them in your ADEPT server (v.4.5.2). If you message me, I can send you a zip of my server.

`python deployment_script.py` or `python redo.py 159`

http://localhost:3000/research-results/open-world-adms

You should now see that, largely, the model/target combinations for part 1 score the same between April and February. It looks like Zero shot CR is not always the same.


brianpippin (approved):
Script ran with no issues for me.

Garycheng92 (approved):
Script ran fine for me

PR #249, opened 2026-08-24
ITM-1436: Text-based Configs for Canada Triangle Exp

dereknop:
`python redo.py 160` or `python deployment_script.py`

This creates the text-based configs for the Canadian portion of the triangle experiment. All of the configs are modified versions of the binomial probe sets from the June 2026 eval. See [dashboard pr](https://github.com/NextCenturyCorporation/itm-evaluation-dashboard/pull/520) after the script runs.

### DIFF
--- a/scripts/_1_6_0_add_canada_text_scenarios.py
+++ b/scripts/_1_6_0_add_canada_text_scenarios.py
@@ -8,4 +8,10 @@ def main(mongo_db):
             continue
         del config['_id']
         config['eval'] = 'Phase 2 Canada Evaluation'
+
+        for page in config['pages']:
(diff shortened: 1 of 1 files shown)


nextcen-dgemoets (commented):
According to my Google search, re `June2026-AF-assess`:
- `The Canadian military primarily uses the abbreviation PW instead of POW for prisoners of war in its official modern doctrine.`

I have no idea if this is correct or if this is the term we should use, but perhaps it's worth verifying?

Also re: `June2026-PS-assess`:
- `The Canadian military does not use the Mk 153 Shoulder-launched Multi-purpose Assault Weapon (SMAW), which is primarily a specialty weapon of the United States Marine Corps. Instead, the Canadian Army relies on other portable anti-armor and support weapons.`
  - `Carl Gustaf: An 84mm recoilless multi-role weapon that Canadian troops commonly call the "Carl G"`
  - `M72 LAW: A lightweight, single-use disposable 66mm rocket launcher.`

Maybe we should consider changing this, too.

PR #250, opened 2026-08-25
ITM-1441: KDMA Fetch For Trinary Compatibility Patch

patmole99:
This PR updates the June 2026 repop and probe matcher scripts to support the addition of trinary sessions. It also includes a deployment script that runs the repop script and probe matcher back to back.

To test these changes, make sure you are using the latest version of the ADEPT server. Older versions will not return the expected KDMA values. Also verify that the ADEPT server’s `openapi_server/data/scenario/` directory contains all of the following files:

* `June2026-AF-assess.yaml`
* `June2026-MF-assess.yaml`
* `June2026-PS-assess.yaml`
* `June2026-SS-assess.yaml`
* `June2026-AF-assess-trinary.yaml`
* `June2026-PS-assess-trinary.yaml`
* `june2026-desert-openworld.yaml`
* `june2026-urban-openworld.yaml`

After adding these files, make sure the ADEPT server has been rebuilt so that they are included in the running server.

Once your environment is configured correctly, run:

`py redo.py 160`

This script will take some time to complete. After it finishes, use MongoDB Compass to query the `humanSimulator` collection with `{evalNumber: 17}`. Spot-check the returned documents and confirm that the values in `text_kdmas` are populated and that the values in `alignment_scores` are not empty or `null`.

### DIFF
--- a/june2026_text_repop.py
+++ b/june2026_text_repop.py
@@ -20,7 +20,7 @@ without creating sessions or updating Mongo.
 --timeout <seconds> ADEPT request timeout. Defaults to 120.
 --skip-af-ss Skip AF-SS_sessionId creation/update.
 --skip-trinary Skip trinary combinedSessionId creation/update.
---skip-mf-individual Skip MF individualSessionId creation/update.
+--create-mf-individual Also create the legacy MF individualSessionId.
 
 Examples:
 py june2026_text_repop.py --eval-number 17 --pid 202606112 --dry-run -v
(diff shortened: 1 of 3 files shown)


june2026_text_repop.py
```diff
@@ -528,10 +528,15 @@ def process_participant(
...
             )
 
-            if trinary_sid and created:
+            if trinary_sid and created and trinary_kdmas is None:
+                log(
+                    "  Error: the new combined trinary session did not return a "
+                    "computed KDMA profile. Mongo will not be updated, preventing "
+                    "a new session id from being paired with stale combinedKdmas."
+                )
+            elif trinary_sid and created:
```
Garycheng92: Should there also be this check for combined binomial and AF-SS? They still write new session IDs even if the KDMA fetch fails so there would be stale KDMAs paired with fresh session IDs

nextcen-dgemoets:
This is going to need to change to 163 or even 164 based on #252 and #253.

patmole99:
> Script ran fine-- there were four cases of:
> 
> ```
> No probes available for KDMA calculation.
> No human session_id available for MF alignment.
> No human session_id available for AF alignment.
> ```
> 
> Three out of those four cases (202606128 desert, 202606135 desert+urban, but not 202606134 desert) had `alignment_scores` of `null`. I don't know if this was expected or a problem, but just reporting it. Even in those cases, they had valid `text_kdmas`.
> 
> Thoughts?

Thanks for pointing those out. I went through each of the raw logs for those and they seem to be a result of incomplete or duplicate runs, so I removed them from the repo.

The only other warning is missing text documents for 202606213. Nothing I can really do about this one unless we find the text documents somehow.

patmole99:
> This is going to need to change to 163 or even 164 based on #252 and #253.

Gotcha, I'll take a look at the PRs and repo right before I merge and switch it to whatever number it needs to be. Gotta make sure prod adept is on the latest version before running this, so I may hold off on merging until I confirm that.

dereknop (approved):
Image

All of the KDMAs are populated correctly now and the alignments are present.

nextcen-dgemoets (approved):
Well, I'm guessing that we've addressed all of the JSON/csv files we can.

As I mentioned in #255 , both PRs have several JSON and csv files deleted. To avoid conflicts, perhaps these deletions should be unstaged from one of these PRs.

